PR #10, opened 2021-04-06
Add vsphere StorageClass for our CI

jsafrane:
The operator does not create a default storage class yet. Use a pre-defined one for CI.

In addition, update the manifest to vSphere


bertinatto (commented):
/lgtm

PR #11, opened 2021-04-06
Add storage class datastore url

gnufied:
Create datastore url based storageclass.


pkg/operator/storageclasscontroller/storageclasscontroller.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+	return factory.New().WithInformers(
+		configMapInformer.Informer(),
+		secretInformer.Informer(),
+		infraInformer.Informer(),
+		operatorClient.Informer(),
+	).WithSync(
+		c.sync,
+	).ResyncEvery(
+		time.Minute, // TODO: figure out what's the idead resync time for this controller
```
bertinatto, line 97: Perhaps bump to 20 min? Not in this PR, but we'll have to bump it the targetconfigcontroller too...

pkg/operator/storageclasscontroller/storageclasscontroller.go
```diff
@@ -0,0 +1,228 @@
...
+	c := &StorageClassController{
+		name:            name,
+		targetNamespace: targetNamespace,
+		manifest:        manifest,
+		kubeClient:      kubeClient,
+		operatorClient:  operatorClient,
+		configMapLister: configMapInformer.Lister(),
+		secretLister:    secretInformer.Lister(),
+		infraLister:     infraInformer.Lister(),
+		recorder:        rc,
```
bertinatto, line 87: nit: we wouldn't need this fiedl, as we can get the recorder from `syncContext`

pkg/operator/storageclasscontroller/storageclasscontroller.go
```diff
@@ -0,0 +1,228 @@
...
+	if err != nil {
+		klog.Errorf("error syncing storage class: %v", err)
+		return err
+	}
+	if _, _, err := v1helpers.UpdateStatus(c.operatorClient,
+		v1helpers.UpdateConditionFn(controllerAvailableCondition),
+	); err != nil {
+		return err
+	}
+	return nil
```
bertinatto, line 140: nit: we could just `return err`

pkg/operator/storageclasscontroller/storageclasscontroller.go
```diff
@@ -0,0 +1,228 @@
...
+		configMapInformer.Informer(),
+		secretInformer.Informer(),
+		infraInformer.Informer(),
+		operatorClient.Informer(),
+	).WithSync(
+		c.sync,
+	).ResyncEvery(
+		time.Minute, // TODO: figure out what's the idead resync time for this controller
+	).WithSyncDegradedOnError(
+		operatorClient,
```
bertinatto, line 99: Do we really want to degrade the operator if we can't create the SC?

We could take a conservative approach for this operator and don't degrade. At least until we are more familiar how this operator behaves on the field.

pkg/operator/storageclasscontroller/storageclasscontroller.go
```diff
@@ -0,0 +1,228 @@
...
+	cfg := new(vsphere.VSphereConfig)
+	err = gcfg.ReadStringInto(cfg, cfgString)
+	if err != nil {
+		return "", err
+	}
+
+	username, password, err := c.getCredentials(cfg)
+	if err != nil {
+		klog.Errorf("error getCredentials: %v", err)
+		return "", fmt.Errorf("failed to get credentials: %w", err)
```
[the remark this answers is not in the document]
gnufied: I have tried to be somewhat verbose on purpose with logging. This controller is not logging too much anyhow. but some of the duplicate messages can be deleted.

pkg/operator/storageclasscontroller/storageclasscontroller.go
gnufied: hmm, is there a reason you think  - creating storageclass would fail in a healthy cluster? With storage policy, it was bit more questionable because we needed more permissions etc. But with this code, I think degrading cluster is fine.

pkg/operator/storageclasscontroller/storageclasscontroller.go
```diff
@@ -0,0 +1,228 @@
...
+		return err
+	}
+
+	err = c.syncStorageClass(ctx, datastoreURL)
+	if err != nil {
+		klog.Errorf("error syncing storage class: %v", err)
+		return err
+	}
+	if _, _, err := v1helpers.UpdateStatus(c.operatorClient,
+		v1helpers.UpdateConditionFn(controllerAvailableCondition),
```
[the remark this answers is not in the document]
gnufied: I am not entirely sure, but in other places where controllers are simpler - we didn't add `Progressing` condition.

pkg/operator/storageclasscontroller/storageclasscontroller.go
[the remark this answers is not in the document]
gnufied: removed most of the extra logs.

pkg/operator/storageclasscontroller/storageclasscontroller.go
```diff
@@ -0,0 +1,228 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to access datastore %s: %s", dsName, err)
+	}
+
+	var dsMo mo.Datastore
+	pc := property.DefaultCollector(dc.Client())
+	properties := []string{DatastoreInfoProperty, SummaryProperty}
+	err = pc.RetrieveOne(ctx, ds.Reference(), properties, &dsMo)
+	if err != nil {
+		return &dsMo, fmt.Errorf("error getting properties of datastore %s: %v", dsName, err)
```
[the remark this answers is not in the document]
gnufied: fixed

pkg/operator/storageclasscontroller/storageclasscontroller.go
gnufied: I changed this to 5 minute, in case user manually changes storageclass.

assets/storageclass.yaml
```diff
@@ -0,0 +1,9 @@
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: standard-csi
```
bertinatto, line 4: I picked a random name here. Is there a better name?

pkg/operator/storageclasscontroller/storageclasscontroller.go
```diff
@@ -0,0 +1,215 @@
...
+		operatorClient:  operatorClient,
+		configMapLister: configMapInformer.Lister(),
+		secretLister:    secretInformer.Lister(),
+		infraLister:     infraInformer.Lister(),
+	}
+	return factory.New().WithInformers(
+		configMapInformer.Informer(),
+		secretInformer.Informer(),
+		infraInformer.Informer(),
+		operatorClient.Informer(),
```
bertinatto, line 85: What do you think about adding a StorageClass informer here as well, so that this controllers resyncs when something changes a StorageClass?

pkg/operator/storageclasscontroller/storageclasscontroller.go
bertinatto: Ack. I was more concerned about a hypothetical use-case where the cluster doesn't have any datastore. I'm far from being familiar with vSphere, so I don't even know if that's possible.

I'm fine with degrading the operator if such scenario is unlikely to exist.

pkg/operator/storageclasscontroller/storageclasscontroller.go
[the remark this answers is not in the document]
bertinatto: Other CSI operators use the static resource controller to create the SC, and that controller only sets the `Degraded` condition:

https://github.com/bertinatto/vmware-vsphere-csi-driver-operator/blob/f41843e4353821f41a50831d5592ab1541a81002/vendor/github.com/openshift/library-go/pkg/operator/staticresourcecontroller/static_resource_controller.go#L169-L211

Should we do the same?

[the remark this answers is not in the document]
bertinatto: hm, is there any chance the API calls can hang for a while (for _reasons_)? If so, perhaps setting `Progressing` isn't a bad idea after all...

assets/storageclass.yaml
gnufied: Looks like same name is being used by other CSI drivers too. So may be we should stick to it.

pkg/operator/storageclasscontroller/storageclasscontroller.go
gnufied: We could, but we only created namespaced informers in this operator. We could do this in a follow up.

assets/storageclass.yaml
bertinatto: The notation I used in other CSI SCs was: _default_sc_name_used_intree + "-csi"_.

`standard` is the default SC for GCE I think (where this was probably copied from). I think Cinder copied GCE too.

I'm fine with this, but I'd be nice to use the same name as the default SC for intree plugin and append `-csi`.

gnufied: hmm, that would logically lead to `thin-csi` . I think there are no better names as such here. I have no strong opinions on this one. Do you prefer `thin-csi` ? I could name it `thin-csi`.

gnufied: I renamed it to `thin-csi`

pkg/operator/storageclasscontroller/storageclasscontroller.go
bertinatto: ack, I'll add it to my TODO

bertinatto (commented):
/lgtm

PR #12, opened 2021-04-06
Disable unsupported features

jsafrane:
Snapshots and expansion are not supported by the CSI driver yet.

@openshift/storage


bertinatto (commented):
/lgtm

PR #13, opened 2021-04-08
Resync controller when user changes StorageClass

bertinatto:
CC @openshift/storage

### DIFF
--- a/pkg/operator/storageclasscontroller/storageclasscontroller.go
+++ b/pkg/operator/storageclasscontroller/storageclasscontroller.go
@@ -34,7 +34,7 @@ const (
 	cloudConfigNamespace  = "openshift-config"
 	DatastoreInfoProperty = "info"
 	SummaryProperty       = "summary"
-	resyncDuration        = 5 * time.Minute
+	resyncDuration        = 20 * time.Minute
 )
 
 type StorageClassController struct {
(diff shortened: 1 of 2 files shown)


pkg/operator/storageclasscontroller/storageclasscontroller.go
```diff
@@ -83,6 +84,9 @@ func NewStorageClassController(
 		secretInformer.Informer(),
 		infraInformer.Informer(),
 		operatorClient.Informer(),
+		// This informer isn't used anywhere else in the code, but it's added here
+		// to make this controller resyncs when users change StorageClasses.
+		scInformer.Informer(),
 	).WithSync(
```
gnufied, line 90: We could increase the resync interval to 20 minutes, now that we are watching storage classes too.

gnufied:
Just putting on hold for a bit, so as fabio can address the review comment.

/hold

pkg/operator/storageclasscontroller/storageclasscontroller.go
bertinatto: I was going to do that initially, but I left as is because both CSI controllers resync every minute.

bertinatto: Set to 20 min.

PR #19, opened 2021-04-13
Bug 1948415: Set the log level in operands

bertinatto:
Sidecar containers have the log level set through the `${LOG_LEVEL}` placeholder, as usual. That placeholder doesn't work for vSphere CSI driver and syncer, though.

First of all, the Deployment controller from library-go replaces `${LOG_LEVEL}` with a _number_ representing the log level. However, both the CSI driver and syncer containers expect the log level to be set through the environment variable `X_CSI_LOG_LEVEL` as a **string**, as in Debug, Trace, etc.

Another problem is that the CSI driver uses logrus instead of klog; and numbers representing log levels have different meanings. For instance, the verbosity value for Debug in klog is 4, while in logrus it's 5.

In order to overcome these differences, this patch takes the following approach: when CR.Spec.LogLevel is Debug or higher, we set the `X_CSI_DEBUG` environment variable in the CSI driver and syncer to a positive value.

CC @openshift/storage


pkg/operator/starter.go
```diff
@@ -220,7 +221,29 @@ func WithSyncerImageHook() csidrivercontrollerservicecontroller.DeploymentHookFu
 	}
 }
 
-func newEnvVar(secretName, envVarName, key string) v1.EnvVar {
+// WithLogLevelDeploymentHook sets the X_CSI_DEBUG environment variable to a positive
+// value when CR.LogLevel is Debug or higher.
+func WithLogLevelDeploymentHook() csidrivercontrollerservicecontroller.DeploymentHookFunc {
```
jsafrane, line 226: I don't see it used anywhere. In addition, should there be WithLogLevelDaemonSetHook?

bertinatto: Sorry, I forgot to mark the PR as WIP

go.mod
```diff
@@ -4,16 +4,26 @@ go 1.16
...
 	github.com/spf13/cobra v1.1.1
 	github.com/spf13/pflag v1.0.5
 	github.com/vmware/govmomi v0.23.1
+	go.uber.org/multierr v1.6.0 // indirect
+	go.uber.org/zap v1.16.0 // indirect
+	golang.org/x/tools v0.0.0-20200825202427-b303f430e36d // indirect
+	google.golang.org/appengine v1.6.6 // indirect
+	google.golang.org/genproto v0.0.0-20201112120144-2985b7af83de // indirect
 	gopkg.in/gcfg.v1 v1.2.3
+	honnef.co/go/tools v0.0.1-2020.1.4 // indirect
```
bertinatto, line 26: This is the result of `go mod vendor && go mod tidy`. I probably forgot to clean this up in my initial PR.

pkg/operator/starter.go
bertinatto: Fixed

pkg/generated/bindata.go
```diff
@@ -290,7 +290,7 @@ metadata:
     csi.openshift.io/managed: "true"
 spec:
   attachRequired: true
-  podInfoOnMount: true
+  podInfoOnMount: false
 `)
```
gnufied, line 294: Did we intend to change this? I don't see corresponding change in YAML.

bertinatto: I guess whoever updated `csidriver.yaml` last time didn't run `make update`.

PR #22, opened 2021-04-27
Bug 1953810: Update CSI driver configuration to the latest version

gnufied:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1953810

After these fixes, we no longer see issues we saw in - https://github.com/openshift/vmware-vsphere-csi-driver-operator/pull/9 and using storage policy works as expected.

### DIFF
--- a/assets/vsphere_features_config.yaml
+++ b/assets/vsphere_features_config.yaml
@@ -1,6 +1,8 @@
 apiVersion: v1
 data:
   "csi-migration": "false" # csi-migration feature is only available for vSphere 7.0U1
+  "csi-auth-check": "true"
+  "online-volume-extend": "true"
 kind: ConfigMap
 metadata:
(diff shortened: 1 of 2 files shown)


assets/vsphere_features_config.yaml
```diff
@@ -1,6 +1,8 @@
 apiVersion: v1
 data:
   "csi-migration": "false" # csi-migration feature is only available for vSphere 7.0U1
+  "csi-auth-check": "true"
+  "online-volume-extend": "true"
```
bertinatto, line 5: https://github.com/kubernetes-sigs/vsphere-csi-driver/blob/d73639381c899762c715d489a7a6a732357af1e2/pkg/csi/service/common/constants.go#L224-L247

bertinatto (commented):
/lgtm

PR #24, opened 2021-05-10
Bug 1954671: disable volume expansion support in vsphere csi driver

dobsonj:


assets/storageclass.yaml
```diff
@@ -6,4 +6,4 @@ provisioner: csi.vsphere.vmware.com
 parameters:
   datastoreurl: "${DATASTORE_URL}"
 volumeBindingMode: WaitForFirstConsumer
-allowVolumeExpansion: true
+allowVolumeExpansion: false
```
bertinatto, line 9: It could be interesting to add a comment stating why this is currently disabled with a link to the ticket/issue that needs to be fixed.

bertinatto: Also, we need to run `make update` to update the bindata file and include that in this PR.

gnufied: may be we need a verify job to catch failure to run `make update`  somewhere

dobsonj:
Verified that after deploying this change, resize fails with an appropriate error message:
```
$ oc patch pvc claim1 -p '{"spec": {"resources": {"requests": {"storage":"2Gi"} } } }'
Error from server (Forbidden): persistentvolumeclaims "claim1" is forbidden: only dynamically provisioned pvc can be resized and the storageclass that provisions the pvc must support resize
```

assets/storageclass.yaml
jingxu97: actually resize only support block volume, not vSAN, I think.

gnufied: hmm, so I think at some point this document was changed. But I could confirm with vmware engineers again. In any case - volume expansion is still beta and hence it makes sense for us to keep it disabled until we could test and validate it.

PR #28, opened 2021-06-18
Add cluster CA bundle

jsafrane:
The CA bundle may contain certificate used to connect to vCenter

Related to https://bugzilla.redhat.com/show_bug.cgi?id=1969719, however, this is about the CSI driver, not vsphere-problem-detector and thus may not be necessary to be backported.

/assign @gnufied

### DIFF
--- a/assets/ca_configmap.yaml
+++ b/assets/ca_configmap.yaml
@@ -1,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  labels:
+    # This label ensures that the OpenShift Certificate Authority bundle
+    # is added to the ConfigMap.
+    config.openshift.io/inject-trusted-cabundle: "true"
+  name: vsphere-csi-driver-trusted-ca-bundle
+  namespace: openshift-cluster-csi-drivers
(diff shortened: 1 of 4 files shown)


assets/ca_configmap.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  labels:
+    # This label ensures that the OpenShift Certificate Authority bundle
+    # is added to the ConfigMap.
+    config.openshift.io/inject-trusted-cabundle: "true"
+  name: vsphere-csi-driver-trusted-ca-bundle
+  namespace: openshift-cluster-csi-drivers
```
bertinatto, line 9: Who is going to apply this ConfigMap?

gnufied: I think some cluster component populates this configmap. I don't know which one though.

bertinatto: The network operator populates this ConfigMap once it exists in the cluster, but first something must create it, right? I don't see this operator creating/applying it.

jsafrane: Good catch! Added to `WithStaticResourcesController()`

bertinatto (commented):
/lgtm

PR #31, opened 2021-07-13
Add code to create storage policy with cluster-id

gnufied:
Adds code to create storage policy with cluster-id


pkg/operator/storageclasscontroller/vmware.go
```diff
@@ -4,20 +4,56 @@ import (
...
 	secretName = "vmware-vsphere-cloud-credentials"
+	apiTimeout = 10 * time.Minute
+
+	categoryNameTemplate = "openshift-%s"
+	policyNameTemplate   = "openshift-storage-policy-%s"
+	vim25Prefix          = "urn:vim25:"
 )
 
-func (c *StorageClassController) newClient(ctx context.Context, cfg *vsphere.VSphereConfig, username, password string) (*govmomi.Client, error) {
+var associatedTypesRaw = []string{"StoragePod", "Datastore", "ResourcePool", "VirtualMachine"}
```
jsafrane, line 36: If I understand it correctly, this is list if types that the *operator* needs to add to a category created by the installer. Should it contain `VirtualMachine` and `ResourcePool`?

If it's the full list of types that OCP may need, then `Folder` is missing.

pkg/operator/storageclasscontroller/vmware.go
```diff
@@ -26,14 +62,248 @@ func (c *StorageClassController) newClient(ctx context.Context, cfg *vsphere.VSp
...
+		}
+		catId, err := tagManager.CreateCategory(ctx, category)
+		if err != nil {
+			return fmt.Errorf("error creating category %s: %v", v.categoryName, err)
+		}
+		category.ID = catId
+	} else {
+		existingAssociatedTypes := category.AssociableTypes
+		associatedTypes = updateAssociatedTypes(existingAssociatedTypes)
+		klog.Infof("existing categories are: %+v", existingAssociatedTypes)
```
jsafrane, line 180: `V(4)`?

pkg/operator/storageclasscontroller/vmware.go
```diff
@@ -26,14 +62,248 @@ func (c *StorageClassController) newClient(ctx context.Context, cfg *vsphere.VSp
...
+		category.AssociableTypes = associatedTypes
+		klog.Infof("final categories are: %+v", associatedTypes)
+		err := tagManager.UpdateCategory(ctx, category)
```
jsafrane, line 183: Please log that category has been updated

pkg/operator/storageclasscontroller/vmware.go
```diff
@@ -26,14 +62,248 @@ func (c *StorageClassController) newClient(ctx context.Context, cfg *vsphere.VSp
...
+
+	associatedTypes := appendPrefix(associatedTypesRaw)
+	if category == nil || category.ID == "" {
+		category = &tags.Category{
+			Name:            v.categoryName,
+			Description:     "Container Orchestrator that uses this datastore",
+			AssociableTypes: associatedTypes,
+			Cardinality:     "SINGLE",
+		}
+		catId, err := tagManager.CreateCategory(ctx, category)
```
jsafrane, line 172: Please log that category has been created

pkg/operator/storageclasscontroller/vmware.go
```diff
@@ -26,14 +62,248 @@ func (c *StorageClassController) newClient(ctx context.Context, cfg *vsphere.VSp
...
+		tag = &tags.Tag{
+			Name:        v.tagName,
+			Description: "Datastore is used by openshift",
+			CategoryID:  category.ID,
+		}
+		tagID, err := tagManager.CreateTag(ctx, tag)
+		if err != nil {
+			return fmt.Errorf("error creating tag %s: %v", v.tagName, err)
+		}
+		tag.ID = tagID
```
jsafrane, line 203: Please log that a tag has been created.
Again, should this tag be created by the installer? A warning would be appropriate if so.

pkg/operator/storageclasscontroller/vmware.go
```diff
@@ -26,14 +62,248 @@ func (c *StorageClassController) newClient(ctx context.Context, cfg *vsphere.VSp
...
+	} else if tag.CategoryID != category.ID {
+		tag = &tags.Tag{
+			Name:        v.tagName,
+			Description: "Datastore is used by openshift",
+			CategoryID:  category.ID,
+			ID:          tag.ID,
+		}
+		err := tagManager.UpdateTag(ctx, tag)
```
jsafrane, line 211: Please log that a tag has been updated

pkg/operator/storageclasscontroller/vmware.go
```diff
@@ -26,14 +62,248 @@ func (c *StorageClassController) newClient(ctx context.Context, cfg *vsphere.VSp
...
+		if err != nil {
+			return fmt.Errorf("error updating tag %s: %v", v.tagName, err)
+		}
+	}
+
+	dsName := v.connection.config.Workspace.DefaultDatastore
+	err = tagManager.AttachTag(ctx, tag.ID, ds)
```
jsafrane, line 218: What if the tag is already attached from previous sync? Will it return error or is it basically a noop?

pkg/operator/storageclasscontroller/vmware.go
```diff
@@ -26,14 +62,248 @@ func (c *StorageClassController) newClient(ctx context.Context, cfg *vsphere.VSp
...
+	}
+
+	tag, err := tagManager.GetTag(ctx, v.tagName)
+	if err != nil && !notFoundError(err) {
+		return fmt.Errorf("error finding tag %s: %v", v.tagName, err)
+	}
+	if tag == nil || tag.ID == "" {
+		tag = &tags.Tag{
+			Name:        v.tagName,
+			Description: "Datastore is used by openshift",
```
jsafrane, line 196: "Added by openshift-vsphere-csi-driver-operator do not remove" to match existing tags?

pkg/operator/storageclasscontroller/vmware.go
```diff
@@ -26,14 +62,248 @@ func (c *StorageClassController) newClient(ctx context.Context, cfg *vsphere.VSp
...
+		}},
+	}
+
+	pid, err := pbmClient.CreateProfile(ctx, policySpec)
+	if err != nil {
+		msg := fmt.Sprintf("error creating profile: %v", err)
+		klog.Errorf(msg)
+		return fmt.Errorf(msg)
+	}
+	klog.Infof("Successfully created profile %s", pid.UniqueId)
```
jsafrane, line 268: `V(2)`?

pkg/operator/storageclasscontroller/vmware.go
```diff
@@ -26,14 +62,248 @@ func (c *StorageClassController) newClient(ctx context.Context, cfg *vsphere.VSp
...
+	profiles, err := pbmClient.RetrieveContent(ctx, ids)
+	if err != nil {
+		msg := fmt.Sprintf("error fetching policy profiles: %v", err)
+		klog.Errorf(msg)
+		return false, fmt.Errorf(msg)
+	}
+
+	for _, p := range profiles {
+		if p.GetPbmProfile().Name == v.policyName {
+			klog.Infof("Found existing profile with same name: %s", p.GetPbmProfile().Name)
```
jsafrane, line 302: `V(2)`?

pkg/operator/storageclasscontroller/vmware.go
```diff
@@ -54,3 +324,24 @@ func (c *StorageClassController) getCredentials(cfg *vsphere.VSphereConfig) (str
 
 	return string(username), string(password), nil
 }
+
+func notFoundError(err error) bool {
+	errorString := err.Error()
+	r := regexp.MustCompile("404")
```
jsafrane, line 330: This looks error prone. How does the message look like? Can we at least grep for `^404`?

pkg/operator/storageclasscontroller/vmware.go
gnufied: If we do end up creating the category in rare cases - we are going to need full list. So, I am going to copy `folder` over here.

pkg/operator/storageclasscontroller/vmware.go
gnufied: This is how message looks like btw:

```
func (e *statusError) Error() string {
	return fmt.Sprintf("%s %s: %s", e.res.Request.Method, e.res.Request.URL, e.res.Status)
}
```

Unfortunately vsphere SDK has chosen to obscure error code from users of the library here. It returns following type as an error:

```
type statusError struct {
	res *http.Response
}
```

since only public method that struct has is `Error()` - I can't extract the error code.

pkg/operator/storageclasscontroller/vmware.go
gnufied: it should be NOOP yes. I have tested this by deleting the policy, which runs the whole thing again.

pkg/operator/storageclasscontroller/vmware.go
gnufied: done

pkg/operator/storageclasscontroller/vmware.go
gnufied: done

pkg/operator/storageclasscontroller/vmware.go

pkg/operator/storageclasscontroller/vmware.go
```diff
@@ -26,14 +62,248 @@ func (c *StorageClassController) newClient(ctx context.Context, cfg *vsphere.VSp
...
+	}
+
+	associatedTypes := appendPrefix(associatedTypesRaw)
+	if category == nil || category.ID == "" {
+		category = &tags.Category{
+			Name:            v.categoryName,
+			Description:     "Container Orchestrator that uses this datastore",
+			AssociableTypes: associatedTypes,
+			Cardinality:     "SINGLE",
+		}
```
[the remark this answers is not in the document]

pkg/operator/storageclasscontroller/vmware.go
gnufied: So `404` string will be at the end - may be I can look for `404` towards the end of the string. Not sure if will be that much better though. :(

gnufied: actually - since it prints `Status` and not `StatusCode` - it is even more crazier, because that would print `blah blah - 404 Not Found`

pkg/operator/storageclasscontroller/vmware.go
```diff
@@ -326,6 +331,7 @@ func (c *StorageClassController) getCredentials(cfg *vsphere.VSphereConfig) (str
 }
 
 func notFoundError(err error) bool {
+	// httpError := err.(*http.Response)
```
jsafrane, line 334: can you please remove the commented line?

PR #43, opened 2021-09-13
[wip] substitute datastore path at job runtime based on hosting cloud

rvanderp3:
IBM Cloud is enabled as a possible platform where CI jobs can run.  At present, the datastore URL is hard coded in the storageclass.  The intent of this PR is to templatize the storageclass and allow the datastore path to be provided at job runtime.  Associated PR https://github.com/openshift/release/pull/21819


rvanderp3:
Closing as storage class is no longer in use.

PR #47, opened 2021-10-04
Bug 2009859: Close connection to vCenter API

bertinatto:
Currently, every time the operator syncs a new session is established.

This patch changes the current behavior by closing the session after it was used.

This should help improving the current situation, however we also need to investigate the upstream CSI driver which apparently is leaking sessions as well.

CC @openshift/storage

### DIFF
--- a/pkg/operator/storageclasscontroller/vmware.go
+++ b/pkg/operator/storageclasscontroller/vmware.go
@@ -41,6 +41,7 @@ type vCenterInterface interface {
 	checkForExistingPolicy(ctx context.Context) (bool, error)
 	createOrUpdateTag(ctx context.Context, ds *mo.Datastore) error
 	createStorageProfile(ctx context.Context) error
+	close(ctx context.Context) error
 }
 
 type vCenterAPI struct {
@@ -70,9 +71,11 @@ func newVCenterAPI(ctx context.Context, cfg *vsphere.VSphereConfig, username, pa
 	if err != nil {
 		return nil, err
 	}
+
 	// We also need to authenticate with the restClient
 	restClient := rest.NewClient(client.Client)
 	userInfo := url.UserPassword(username, password)
+
 	err = restClient.Login(ctx, userInfo)
 	if err != nil {
 		msg := fmt.Sprintf("error logging into vcenter: %v", err)
@@ -272,6 +275,10 @@ func (v *vCenterAPI) createStorageProfile(ctx context.Context) error {
 	}
 	klog.V(2).Infof("Successfully created profile %s", pid.UniqueId)
 	return nil
+}
+
+func (v *vCenterAPI) close(ctx context.Context) error {
+	return v.connection.client.Logout(ctx)
 }
 
 func (v *vCenterAPI) checkForExistingPolicy(ctx context.Context) (bool, error) {
--- a/pkg/operator/storageclasscontroller/storageclasscontroller.go
+++ b/pkg/operator/storageclasscontroller/storageclasscontroller.go
@@ -160,11 +160,17 @@ func (c *StorageClassController) syncStoragePolicy(ctx context.Context) (string,
 	if err != nil {
 		return "", fmt.Errorf("failed to get credentials: %v", err)
 	}
-	apiClient, err := newVCenterAPI(ctx, cfg, username, password, infra)
 
+	apiClient, err := newVCenterAPI(ctx, cfg, username, password, infra)
 	if err != nil {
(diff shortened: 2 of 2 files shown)


bertinatto:
CC @rvanderp3

pkg/operator/storageclasscontroller/vmware.go
```diff
@@ -274,6 +277,10 @@ func (v *vCenterAPI) createStorageProfile(ctx context.Context) error {
 	return nil
 }
 
+func (v *vCenterAPI) close(ctx context.Context) error {
+	return v.connection.client.Logout(ctx)
+}
+
```
gnufied, line 283: do we need to also logout from restclient? something like `v.connection.restClient.Logout` ?

gnufied:
@bertinatto did you had a chance to measure session/connection growth after this change?

jcpowermac:
How often is `syncStoragePolicy` being called? Is this going to be a situation of opening and closing the connection?

bertinatto:
> @bertinatto did you had a chance to measure session/connection growth after this change?

@gnufied yes, without this patch the number of open sessions only increase. With this patch, it only increases temporarily until the syn finishes.

pkg/operator/storageclasscontroller/vmware.go
bertinatto: That didn't improve the situation so I decided to leave it out of this PR.

bertinatto:
> How often is `syncStoragePolicy` being called? Is this going to be a situation of opening and closing the connection?

That controller syncs every 20 minutes, or every time there are changes in secrets (openshift-cluster-csi-drivers NS), configMaps (openshift-config NS) and StorageClasses.

jcpowermac:
> > How often is `syncStoragePolicy` being called? Is this going to be a situation of opening and closing the connection?
> 
> That controller syncs every 20 minutes, or every time there are changes in secrets (openshift-cluster-csi-drivers NS), configMaps (openshift-config NS) and StorageClasses.

Thanks for the details, yeah that shouldn't be a problem.

rvanderp3:
> > How often is `syncStoragePolicy` being called? Is this going to be a situation of opening and closing the connection?
> 
> That controller syncs every 20 minutes, or every time there are changes in secrets (openshift-cluster-csi-drivers NS), configMaps (openshift-config NS) and StorageClasses.

During e2e tests, this was fairly frequent(i.e. every several seconds).  Is there something special about e2e tests that would make syncs occur more frequently?

pkg/operator/storageclasscontroller/vmware.go
gnufied: > That didn't improve the situation so I decided to leave it out of this PR.

Is this because we did not measure rest client connections or because logging out from SOAP connection also closes REST connection? Is there a downside of logging from restclient?

gnufied:
> During e2e tests, this was fairly frequent(i.e. every several seconds). Is there something special about e2e tests that would make syncs occur more frequently?

not sure. But may be it is because of those watches. It may be worth looking if we are needlessly updating certain objects. 
So @rvanderp3 @jcpowermac - logging out is an acceptable solution compared to client caching? Logging out is obviously simpler, but it does mean we are creating one or two connections every reconcile interval.

bertinatto:
> > > How often is `syncStoragePolicy` being called? Is this going to be a situation of opening and closing the connection?
> > 
> > 
> > That controller syncs every 20 minutes, or every time there are changes in secrets (openshift-cluster-csi-drivers NS), configMaps (openshift-config NS) and StorageClasses.
> 
> During e2e tests, this was fairly frequent(i.e. every several seconds). Is there something special about e2e tests that would make syncs occur more frequently?

We have a fair number of e2e tests that create and delete special-purpose StorageClasses. This definitely makes the controller re-sync.

I just run our CSI test suite and the number of permanent open sessions haven't increased:

```
[fbertina@ip-10-0-5-26 ~]$ govc session.ls | grep test | awk '{print $7}' | sort | uniq -c
      2 Go-http-client/1.1
      1 govc/0.24.0
      3 k8s-csi-useragent
      1 kubernetes-cloudprovider/v1.21.1+a620f50
      1 machineAPIvSphereProvider
```

gnufied:
I think one pro of this approach here is - if no sync is happening, no persistent connection is held by this operator at all. So in a steady state cluster, it may be preferable.

gnufied:
Okay. lets give this a go and see how it works out. 

/lgtm

PR #51, opened 2021-11-09
[release-4.8] Bug 2021620: Close connection to vCenter API

rvanderp3:
The intention of this PR is to backport a fix to close stale connections which was implemented in #50 and #47 to 4.8.  There are structural differences in pkg/operator/storageclasscontroller/vmware.go which prevent a cherrypick and as such this PR was created.  As is done in prior PRs on this issue, logout is called on the vim25.Client instance which results in the session being logged out and idle connections being closed.


jsafrane:
/cc @chao007 @duanwei33
for cherry-pick-approved label

PR #54, opened 2021-11-16
storageclasscontroller: do not degrade on error

bertinatto:
As we previously discussed, we don't want to degrade the whole cluster if something wrong creating the SC.

CC @openshift/storage

### DIFF
--- a/pkg/operator/storageclasscontroller/storageclasscontroller.go
+++ b/pkg/operator/storageclasscontroller/storageclasscontroller.go
@@ -35,6 +35,9 @@ const (
 	resyncDuration        = 20 * time.Minute
 )
 
+// StorageClassController ultimately creates a StorageClass that
+// uses the vSphere CSI driver to provision volumes. On error,
+// this controller does not set any degraded condition.
 type StorageClassController struct {
 	name            string
 	targetNamespace string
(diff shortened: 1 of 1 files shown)


jsafrane:
The cluster is not degraded, but the operator should also get Disabled, with an error message. Now the error is hidden from user.

jsafrane:
Or maybe not Disabled, but Available: true with the error message. We do it in some operator already.

pkg/operator/storageclasscontroller/storageclasscontroller.go
```diff
@@ -35,6 +35,9 @@ const (
 	resyncDuration        = 20 * time.Minute
 )
 
+// StorageClassController ultimately creates a StorageClass that
+// uses the vSphere CSI driver to provision volumes. On error,
+// this controller does not set any degraded condition.
```
gnufied, line 40: Which bug we are fixing? I thought the proper way of fixing this was to *not* create storageclass if we can't create storageclass. But if cluster meets OCP requirements of VCP CSI driver and if we can't create SC, we *SHOULD* degrade the cluster.

bertinatto: Not a bug, but we discussed this multiple times during our design meetings. Specifically, this was agreed on on Sep 16 but we never created a card for it.

PR #56, opened 2021-11-30
add checks for environment and existing driver

gnufied:
Add checks for environment and existing driver in csi driver operator.


pkg/operator/environmentchecker/environmentcheckcontroller.go
```diff
@@ -0,0 +1,409 @@
...
+		// if we can't fetch CSI driver lets not degrade the cluster, if cluster indeed needs degrading we will catch in
+		// next sync
+		if err != nil {
+			return false
+		}
+		annotations := csiDriver.GetAnnotations()
+		if _, ok := annotations[utils.OpenshiftCSIDriverAnnotationKey]; ok {
+			return true
+		}
+		sc, err := c.scLister.Get(storageClassName)
```
gnufied, line 213: I don't know if we should annotate storageclass we create, so as we can guarantee that it is OCP that created the storageclass. @bertinatto @jsafrane

pkg/operator/environmentchecker/environmentcheckcontroller.go
```diff
@@ -0,0 +1,409 @@
...
+	}
+
+	if c.vSphereConnection != nil {
+		secretConfigHash, err := c.getSecretConfigMapHash(ctx, infra)
+		if err != nil {
+			immediateError = err
+			return
+		}
+
+		if secretConfigHash == c.vSphereConnection.SecretConfigMapHash {
```
gnufied, line 258: cc @derek-pryor @jsafrane @bertinatto this implements a form of connection caching. But the implementation can be tidied up some more.

pkg/operator/environmentchecker/environmentcheckcontroller.go
```diff
@@ -0,0 +1,409 @@
...
+
+	// if operand was not started previously and block upgrade is false and clusterdegrade is also false
+	// then and only then we should start CSI driver operator
+	if !c.operandControllerStarted && !result.BlockUpgrade && !result.ClusterDegrade {
+		go c.runConditionalController(ctx)
+		c.operandControllerStarted = true
+	}
+
+	// sync storage class
+	storageClassSyncResult := c.storageClassController.Sync(ctx, c.vSphereConnection)
```
bertinatto, line 179: I think we shouldn't sync the controller in here, it should run on its own goroutine. For example, if the StorageClass is deleted, only the StorageClass controller should re-sync.

pkg/operator/environmentchecker/environmentcheckcontroller.go
```diff
@@ -0,0 +1,409 @@
...
+	return factory.New().WithInformers(
+		apiClients.OperatorClient.Informer(),
+		configMapInformer.Informer(),
+		apiClients.SecretInformer.Informer(),
+		infraInformer.Informer(),
+		scInformer.Informer(),
+	).WithSync(c.sync).
+		ResyncEvery(resyncDuration).
+		WithSyncDegradedOnError(apiClients.OperatorClient).ToController(c.name, rc)
+}
```
bertinatto, line 120: I'd prefer this controller to stick to checking the environment only, running side-by-side with other controllers. Currently it acts like a manger and IMO it's getting too complex.

If this controller needs to share some of its findings to other controllers, I'd suggest using another custom resource for that (e.g., the ClusterCSIDriver CR).

pkg/operator/environmentchecker/environmentcheckcontroller.go
gnufied: hmm, it is tricky to share vsphere connection between parallel controllers. We will need a global connection pool with caching and everything.

jsafrane (commented):
I did not have time to do a proper review, just posting my nits I found in the first few minutes.

pkg/operator/environmentchecker/environmentcheckcontroller.go
jsafrane: If we want to fix the connection leak(s), we need a small and separate PR that we can backport to 4.9 too.

gnufied: I got rid of connection caching and went for logout at the end of `sync` function approach. I think later approach is superior to connection cache because in that case, we do not hold any connection to vcenter at all.

See attached screenshot - even though our operator does all of its things it holds 0 connections to vcenter. 
[](url)
Screenshot from 2021-12-08 11-05-33

pkg/operator/environmentchecker/environmentcheckcontroller.go
gnufied: We discussed this offline. We can't start CSI controller set controllers if existing driver check failed (otherwise we will have two CSI drivers). Also if we are going to put something in  `ClusterCSIDriver` object for conditional start of CSI controller set, then we will have to wrap CSI controller set in another controller and at that point, I am not sure if this is superior or current approach.

pkg/operator/environmentchecker/driver_starter.go
```diff
@@ -0,0 +1,217 @@
...
+			"rbac/kube_rbac_proxy_binding.yaml",
+			"rbac/prometheus_role.yaml",
+			"rbac/prometheus_rolebinding.yaml",
+			// Static assets used by the webhook
+			// "webhook/config.yaml",
+			// "webhook/sa.yaml",
+			// "webhook/service.yaml",
+			// "webhook/configuration.yaml",
+			// "webhook/rbac/role.yaml",
+			// "webhook/rbac/rolebinding.yaml",
```
bertinatto, line 62: The webhook is being deployed now, so you'd need to uncomment these lines

pkg/operator/environmentchecker/environmentcheckcontroller.go
bertinatto: We did something like this in Manila: the `ManilaController` does a few checks, starts the remaining controllers and sets a `*Disabled` condition. Later on CSO [deals with that condition](https://github.com/bertinatto/cluster-storage-operator/blob/2e76e02cb8479784a582e57a587be8fd5d75cae2/pkg/operator/csidriveroperator/crcontroller.go#L192-L195).

Should we do the same here?

pkg/operator/starter.go
```diff
@@ -70,88 +61,23 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
...
+		KubeClient:      kubeClient,
+		KubeInformers:   kubeInformersForNamespaces,
+		SecretInformer:  secretInformer,
+		NodeInformer:    nodeInformer,
+		ConfigClientSet: configClient,
+		ConfigInformers: configInformers,
+		DynamicClient:   dynamicClient,
+	}
+
+	environmentCheckController := environmentchecker.NewEnvironmentCheckController(
```
bertinatto, line 75: Suggestion: can we call this controller `VSphereController`?

gnufied:
for some reason this PR is showing cached content.

PR #58, opened 2021-12-08
Conditional start vsphere driver

gnufied:
Add checks for environment and existing driver in csi driver operator.


pkg/operator/starter.go
```diff
@@ -70,88 +61,23 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
...
-		nil,
-		WithLogLevelDaemonSetHook(),
-		csidrivernodeservicecontroller.WithObservedProxyDaemonSetHook(),
-	).WithServiceMonitorController(
-		"VMWareVSphereDriverServiceMonitorController",
-		dynamicClient,
-		assets.ReadFile,
-		"servicemonitor.yaml",
-	)
+	commonApiClient := utils.ApiClients{
```
jsafrane, line 64: `commonAPIClient`, `APIClients`

pkg/operator/vspherecontroller/vspherecontroller.go
```diff
@@ -0,0 +1,409 @@
...
+		return true
+	}
+
+	if result.BlockUpgrade {
+		// if we were supposed to only block upgrades but for some reason we have created CSIDriver and storageclass
+		// in past and future syncing fails then we should degrade the cluster
+		csiDriver, err := c.csiDriverLister.Get(utils.VSphereDriverName)
+
+		// if we can't fetch CSI driver lets not degrade the cluster, if cluster indeed needs degrading we will catch in
+		// next sync
```
jsafrane, line 254: The next sync can be quite far away, in 20 minutes in the worst case. IMO degrading the cluster is appropriate here (ie. return err)

pkg/operator/vspherecontroller/vspherecontroller.go
```diff
@@ -0,0 +1,409 @@
...
+	if !c.operandControllerStarted && !degradedOrBlockedFromUpgrades {
+		go c.runConditionalController(ctx)
+		c.operandControllerStarted = true
+	}
+	return nil
+}
+
+func (c *VSphereController) blockUpgradeOrDegradeCluster(ctx context.Context, result checks.ClusterCheckResult) (error, bool) {
+	var clusterCondition string
+	blockUpgrade := c.shouldDegradeCluster(result)
```
jsafrane, line 228: the variable name is misleading

pkg/operator/vspherecontroller/checks/api_interface.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+type KubeApiInterface interface {
+	// ListNodes returns list of all nodes in the cluster.
+	ListNodes() ([]*v1.Node, error)
+	GetCSIDriver(name string) (*storagev1.CSIDriver, error)
+	ListCSINodes() ([]*storagev1.CSINode, error)
+}
+
+type KubeApiInterfaceImpl struct {
```
jsafrane, line 19: `KubeAPIInterfaceImpl`

pkg/operator/vspherecontroller/vspherecontroller.go
```diff
@@ -0,0 +1,409 @@
...
+	if err != nil {
+		return err
+	}
+
+	// if operand was not started previously and block upgrade is false and clusterdegrade is also false
+	// then and only then we should start CSI driver operator
+	if !c.operandControllerStarted && !degradedOrBlockedFromUpgrades {
+		go c.runConditionalController(ctx)
+		c.operandControllerStarted = true
+	}
```
jsafrane, line 222: What happens when all check pass, the conditional controllers start (= the CSI driver controllers are running), a cluster check fails (dunno, a node with HW ver 13 joins the cluster?) and the operator is restarted? Now the CSI driver controllers are not running.

I.e. when a check fails, sometimes the operator syncs the CSI driver Deployment / DaemonSet, and sometimes it does not.

IMO, when (our) CSI driver is present, it should always start the driver controllers, because at some point in the past all checks succeeded and the driver started. The handling should be similar (the same?) as `shouldDegradeCluster`, some refactoring could be helpful.

pkg/operator/vspherecontroller/vspherecontroller.go
```diff
@@ -0,0 +1,409 @@
...
+		CSINodeLister:   c.csiNodeLister,
+		CSIDriverLister: c.csiDriverLister,
+		NodeLister:      c.nodeLister,
+	}
+
+	checkOpts := checks.NewCheckArgs(c.vSphereConnection, checkerApiClient)
+	return c.vSphereChecker.Check(ctx, checkOpts)
+}
+
+func (c *VSphereController) loginToVcenter(ctx context.Context, infra *ocpv1.Infrastructure) checks.ClusterCheckResult {
```
jsafrane, line 303: nit: `loginToVCenter`

pkg/operator/vspherecontroller/vsphere_environment_checker.go
```diff
@@ -0,0 +1,98 @@
...
+		// Maximum interval between checks.
+		Cap: time.Hour * 1,
+	}
+)
+
+type vSphereEnvironmentCheckInterface interface {
+	Check(ctx context.Context, connection checks.CheckArgs) (time.Duration, checks.ClusterCheckResult, bool)
+}
+
+type VsphereEnvironmentChecker struct {
```
jsafrane, line 29: nit: `VSphereEnvironmentChecker`

jsafrane: Maybe I am missing something, but can it be private?

pkg/operator/starter.go
gnufied: fixed.

pkg/operator/vspherecontroller/vspherecontroller.go
gnufied: changed. Also improved the check to check for `CSIDriver` object installed by OCP.

pkg/operator/vspherecontroller/vspherecontroller.go
gnufied: this bit of code was rewritten with different logic. PTAL.

pkg/operator/vspherecontroller/checks/api_interface.go
gnufied: fixed.

pkg/operator/vspherecontroller/vspherecontroller.go
gnufied: Update code to *always* start CSI controller set if we previously created `CSIDriver` object.

pkg/operator/vspherecontroller/vsphere_environment_checker.go
gnufied: made private.

gnufied:
Tested with cluster which had CSI driver installed and verified that something that would only block upgrade, will end up degrading the cluster if CSI driver was installed:


```
  "lastTransitionTime": "2021-12-15T14:45:00Z",
  "message": "node co8-v5sg6-master-1 has hardware version vmx-13, which is below the minimum required version 15",
  "reason": "SyncError",
  "status": "True",
  "type": "VMwareVSphereEnvironmentCheckControllerDegraded" --> I have renamed this controller and pushed the update.
}

```

pkg/operator/storageclasscontroller/storageclasscontroller.go
```diff
@@ -41,147 +27,117 @@ type StorageClassController struct {
...
-	username, password, err := c.getCredentials(cfg)
-	if err != nil {
-		return "", fmt.Errorf("failed to get credentials: %v", err)
+	switch clusterStatus {
+	case checks.ClusterCheckDegrade:
+		degradedMessage := fmt.Sprintf("Marking cluster degraded because %s", lastCheckResult.Reason)
+		klog.Warningf(degradedMessage)
+		c.recorder.Warningf(string(lastCheckResult.CheckStatus), "Marking cluster degraded because %s", lastCheckResult.Reason)
+		degradedCondition.Status = operatorapi.ConditionTrue
+		degradedCondition.Message = degradedMessage
```
jsafrane, line 112: Only `lastCheckResult.Reason` should be in the condition message, "Marking cluster degraded because %s" is implied by the condition name.

(Applies to all conditions below)

pkg/operator/vspherecontroller/vspherecontroller.go
```diff
@@ -0,0 +1,403 @@
...
+		klog.V(4).Infof("%s: vcenter-csi logging out from vcenter", c.name)
+		err := c.vSphereConnection.Logout(ctx)
+		if err != nil {
+			klog.Errorf("%s: error closing connection to vCenter API: %v", c.name, err)
+		}
+	}()
+
+	// if there was an OCP error we should degrade the cluster or if we previously created CSIDriver
+	// but we can't connect to vcenter now, we should also degrade the cluster
+	err, degradedOrBlockedFromUpgrades := c.blockUpgradeOrDegradeCluster(ctx, connectionResult, infra)
```
jsafrane, line 188: `blockUpgradeOrDegradeCluster` is called a lot in this function and it will always update the ClusterCSIDriver conditions.

* This is quite aggressive.
* When e.g.  `loginToVCenter` checks succeeds, the conditions get cleared, but when a subsequent check fails, they get filed again - they will be flipping too much IMO.

Should there be a single function that does `sync()` and returns the final conditions that are applied only once?

gnufied: hmm, if ` loginToVCenter` check succeeds the conditions are not touched at all. Conditions are only updated if either `blockUpgrade` is true.

gnufied: fixed.

pkg/operator/storageclasscontroller/storageclasscontroller.go
gnufied: fixed.

bertinatto (commented):
/lgtm

gnufied:
The flakes seems to be VMC related.

/retest

PR #65, opened 2022-01-11
Bug 2038389: Avoid event spam for same events

gnufied:
Look for existing condition before emitting an event. 

xref https://bugzilla.redhat.com/show_bug.cgi?id=2038389


pkg/operator/vspherecontroller/vspherecontroller.go
```diff
@@ -390,7 +392,7 @@ func (c *VSphereController) updateConditions(ctx context.Context, name string, l
 	if lastCheckResult.BlockUpgrade {
 		blockUpgradeMessage := fmt.Sprintf("Marking cluster un-upgradeable because %s", lastCheckResult.Reason)
 		klog.Warningf(blockUpgradeMessage)
-		c.eventRecorder.Warningf(string(lastCheckResult.CheckStatus), "Marking cluster un-upgradeable because %s", lastCheckResult.Reason)
+		c.meteredEventEmitter.Warn(lastCheckResult)
```
jsafrane, line 395: IMO it would be enough to emit the event only when `allowUpgradeCond` actually changes the existing condition. Would it be better to pass it here (and `Get` it in the caller) instead of the whole `warningEventEmitter`?

gnufied: okay done. PTAL.

pkg/operator/vspherecontroller/vsphere_environment_checker_test.go
```diff
@@ -0,0 +1,126 @@
...
+			if result.CheckStatus != test.result {
+				t.Fatalf("expected test status to be %s, got %s", test.result, result.CheckStatus)
+			}
+			if checkRan != test.checksRan {
+				t.Fatalf("expected checkRan to be %v got %v", test.checksRan, checkRan)
+			}
+			expectedDelay := test.delayFunc()
+			if !compareTimeDiffWithinTimeFactor(expectedDelay, delay) {
+				t.Fatalf("expected delay to %v, got %v", expectedDelay, delay)
+			}
```
jsafrane, line 111: Since this is an unit test, it could probably check `v.backoff.Steps == defaultBackoff.Steps - 2`
(or something like that). IMO anything based on an actual wall clock time is fragile in unit tests.

jsafrane: Btw, it's not really related to the bug that's being fixed here. I'm fine if it's in this PR, but it could be a separate PR too.

gnufied: Pushed a fix for this.

PR #66, opened 2022-01-11
Bug 2039406: Also logout from restclient

gnufied:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2039406

Note to reviewers: While I tested this PR with real cluster and did not see any problems, for me - currently there is no way to measure rest session count, so I don't really know if it *really* works. I do know that it does not break.


jsafrane:
/retest

This looks serious:
```
    --- FAIL: TestSync/when_all_configuration_is_right (4.18s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
```

gnufied:
> This looks serious:

Yep sorry about that fixed.

PR #69, opened 2022-01-17
Bug 2038934: Add custom CA bundle support

derek-pryor:
In OCP, it's possible to set a cluster-wide proxy that operators can use to communicate with external APIs [1]. In addition to that, users can also define a custom CA bundle to be used together with the proxy setting.

Currently, we are not using that custom CA bundle and this patch fixes that. The main benefit of injecting the CA bundle using the CSI controllers, as opposed to statically mount it using the manifest file, is that we can watch for changes in the trusted CA bundle store and roll out a new Deployment with the new CA bundle injected.

[1] https://docs.openshift.com/container-platform/4.10/networking/enable-cluster-wide-proxy.html

CC @openshift/storage


jsafrane:
The operator already has some support for `vsphere-csi-driver-trusted-ca-bundle` ConfigMap:

* It's already used in controller.yaml: https://github.com/openshift/vmware-vsphere-csi-driver-operator/blob/cb1e1c515c263e0693c49d2c84fbf4d0f03ba56f/assets/controller.yaml#L283-L288
* It's already used in node.yaml: https://github.com/openshift/vmware-vsphere-csi-driver-operator/blob/cb1e1c515c263e0693c49d2c84fbf4d0f03ba56f/assets/node.yaml#L148-L153

Both usages are quite static. When the ConfigMap is updated with a new CA bundle, it's injected into the driver Pods (that's what ConfigMaps do), but nobody knows where is the CA bundle cached and our new `WithCABundleDeploymentHook` / `WithCABundleDaemonSetHook`  will make the pods restart and we can't be sure the new pods will use the fresh bundle.

And when the CA bundle is empty or missing, the new hooks will not add the bundle to the Pods. It's unlikely to happen in a real cluster, still, it's better to be robust.

Therefore please remove traces of the ConfigMap volumes from controller.yaml + node.yaml, the hooks will add it there dynamically.

derek-pryor:
> Therefore please remove traces of the ConfigMap volumes from controller.yaml + node.yaml, the hooks will add it there dynamically.

Thanks for catching that. Removed those references.

PR #72, opened 2022-01-28
Bug 2047779: Never degrade cluster on storage policy syncResult

gnufied:
Only degrade if storage class creation fails because of API error

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2047779


jsafrane:
/override ci/prow/e2e-vsphere
I can see the job succeeded, only etcd alerted on slow disk. This PR does not affect it.

PR #77, opened 2022-02-28
Bug 2053104: Do not cache node check results between runs

gnufied:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2053104


pkg/operator/vspherecontroller/checks/check_nodes.go
```diff
@@ -169,6 +169,7 @@ func (n *NodeChecker) Check(ctx context.Context, checkOpts CheckArgs) []ClusterC
 
 	// Map of host names and their ESXI versions
 	n.hostESXIVersions = map[string]bool{}
+	n.results = []ClusterCheckResult{}
```
gnufied, line 172: The only code change that is needed for the fix. rest of is just tests and stuff.

dobsonj:
Code changes look ok to me, just waiting until tests pass to lgtm. The verify job is reporting a gofmt error.

PR #79, opened 2022-03-08
Set fsGroupPolicy in CSIDriver

jsafrane:
And other CSIDriver fields. Since volumes of this CSI driver always support `fsGroup`, don't depend on Kubernetes heuristics (`ReadWriteOnceWithFSType`) and always apply it.

This updates library-go to https://github.com/openshift/library-go/pull/1324 to re-create driver's CSIDriver instance if the operator needs to change it.

It's somewhat related to https://bugzilla.redhat.com/show_bug.cgi?id=2058626, it makes sure the BZ never happens again.

cc @openshift/storage


bertinatto (commented):
/lgtm

PR #81, opened 2022-03-18
Bug 2059791: [vSphere CSI driver Operator] didn't update 'vsphere_csi_driver_error' metric value when fixed the error manually

RomanBednar:
If operator creates a metric indicating error the metric is never
removed and reported even if user resolved the issue.

For example setting incorrect password for vsphere connection degrades
the operator state and emits a metric when periodic sync is performed.

If user then corrects the password and vsphere client succeeds this
metric is persisted.

This patch should ensure we reset the metric if we don't detect any
known negative state (defined in blockUpgradeOrDegradeCluster())

### DIFF
--- a/pkg/operator/vspherecontroller/vspherecontroller.go
+++ b/pkg/operator/vspherecontroller/vspherecontroller.go
@@ -277,8 +277,10 @@ func (c *VSphereController) blockUpgradeOrDegradeCluster(
 		utils.InstallErrorMetric.WithLabelValues(string(result.CheckStatus), clusterCondition).Set(1)
 		updateError := c.updateConditions(ctx, c.name, result, status, operatorapi.ConditionFalse)
 		return updateError, true
+	default:
+		utils.InstallErrorMetric.Reset()
+		return nil, false
 	}
-	return nil, false
 }
 
 func (c *VSphereController) runConditionalController(ctx context.Context) {


RomanBednar:
This is how I verified the patch, another option would be to set the metric to 0 (suggested by @jsafrane ) instead of resetting it, but I was not sure how to deal with the labels properly.

Observe connection failure due to wrong password

```
$ oc exec -n openshift-monitoring statefulsets/prometheus-k8s -- curl -ks -H "Authorization: Bearer $token" https://prometheus-k8s:9091/api/v1/query\?query\=vsphere_csi_driver_error |jq .
{
  "status": "success",
  "data": {
    "resultType": "vector",
    "result": [
      {
        "metric": {
          "__name__": "vsphere_csi_driver_error",
          "condition": "degraded",
          "container": "vmware-vsphere-csi-driver-operator",
          "endpoint": "vsphere-omp",
          "failure_reason": "vsphere_connection_failed",
          "instance": "10.130.0.58:8445",
          "job": "vmware-vsphere-csi-driver-operator-metrics",
          "namespace": "openshift-cluster-csi-drivers",
          "pod": "vmware-vsphere-csi-driver-operator-677c84d57c-kgvwh",
          "service": "vmware-vsphere-csi-driver-operator-metrics"
        },
        "value": [
          1647607903.866,
          "1"
        ]
      }
    ]
  }
}
```

Fix the password

```
$  oc -n openshift-cluster-csi-drivers patch secret/vmware-vsphere-cloud-credentials -p '{"data":{"vcenter.sddc-44-236-21-251.vmwarevmc.com.password":"****************"}}'
secret/vmware-vsphere-cloud-credentials patched
```

Verify the metric has been reset

```
$ oc exec -n openshift-monitoring statefulsets/prometheus-k8s -- curl -ks -H "Authorization: Bearer $token" https://prometheus-k8s:9091/api/v1/query\?query\=vsphere_csi_driver_error |jq .
{
  "status": "success",
  "data": {
    "resultType": "vector",
    "result": []
  }
}
```

gnufied:
Not ideal, but not end of the world either I think.

jsafrane:
The metrics for "fixed" problems will get missing. @gnufied is it a problem? Will any alerting rules remember the old value ("1") and keep alerting?

PR #82, opened 2022-04-08
Bug 2072139: Create separate controller and node roles

jsafrane:
Create a separate Role + ClusterRole + their Bindings for the node ServiceAccount.

~And remove `privileged` role from the controller.~

Reuse `vmware-vsphere-csi-driver-role` / `vmware-vsphere-csi-driver-binding` for the controller ServiceAccount. I am not renaming the object to make sure we do not leak unused ClusterRole/Binding during update.

I used upstream node Role and ClusterRole from
https://github.com/kubernetes-sigs/vsphere-csi-driver/blob/v2.5.1/manifests/vanilla/vsphere-csi-driver.yaml.
`


jsafrane:
/hold
I still need to review if the controller needs to run privileged pods, it looks suspicious

jsafrane:
cc @openshift/storage

bertinatto (commented):
/lgmt

PR #84, opened 2022-04-21
Bug 2076637: Scrape CSI driver + syncer metrics

jsafrane:
The CSI driver does not have configurable port for the metrics and due to `hostNetwork` they will be exposed on the node.

I opened https://github.com/kubernetes-sigs/vsphere-csi-driver/issues/1720 upstream.


jsafrane:
cc @openshift/storage

bertinatto (commented):
/lgtm

PR #87, opened 2022-05-26
Bug 2089419: Don't block upgrades when another CSI driver is found

jsafrane:
The first commit refactors two bools into enum, to be able to add a third possible outcome of a failed check.
The second commit then blocks CSI driver installation when the driver is already installed and it's not the OCP one. The operator looks "Available: true" + "Upgradeable: true", however, the Upgradeable condition has a message `found existing unsupported csi.vsphere.vmware.com driver`.

WIP:
* Add unit test.
* Think about useful alert.

@openshift/storage

### DIFF
--- a/pkg/operator/vspherecontroller/checks/check_error.go
+++ b/pkg/operator/vspherecontroller/checks/check_error.go
@@ -2,6 +2,7 @@ package checks
 
 import (
 	"fmt"
+
 	"github.com/openshift/vmware-vsphere-csi-driver-operator/pkg/operator/utils"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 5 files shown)


pkg/operator/vspherecontroller/checks/check_error.go
```diff
@@ -124,6 +147,11 @@ func CheckClusterStatus(result ClusterCheckResult, apiDependencies KubeAPIInterf
 		}
 
 		return ClusterCheckBlockUpgrade, result
+
+	case CheckActionBlockInstall:
+		return ClusterCheckBlockInstall, result
```
gnufied, line 152: `ClusterCheckBlockInstall` should may be renamed to `ClusterCheckBlockDriverInstall`. I know it is bit mountful but perhaps more accurate.

gnufied, line 152: `ClusterCheckBlockInstall` should may be renamed to `ClusterCheckBlockDriverInstall`. I know it is bit mountful but perhaps more accurate.

jsafrane: done

jsafrane: done

jsafrane:
/retest
installation flakes

PR #90, opened 2022-06-09
Bug 2095248: Report max. nr. of attachable volumes per node

jsafrane:
`"59"` comes from https://github.com/kubernetes-sigs/vsphere-csi-driver/blob/532a1ccb538621261c69abec0043284fb54177b8/pkg/csi/service/node.go#L40

Presence of the the env. variable ensures that the driver reports attach limits in `NodeGetInfo()` CSI call.
cc @openshift/storage


bertinatto (commented):
/lgtm

PR #92, opened 2022-06-27
Bug 2081557: Run at least one replica of Webhook

jsafrane:
There should be always at least one replica of the webhook always running to make sure API requests are processed + the operator does not go `Available: "false"`.

* Add PodDisruptionBudget with 1 replica.
* Use `MaxUnavailable: 1`.
* Tune nr. of replicas to `1` on single-node cluster, `2` everywhere else.
* Run the webhook on regular nodes in hypershift (masters are not there)

All this code is already used in other CSI driver Deployments.

cc @openshift/storage


bertinatto (commented):
/lgtm

PR #93, opened 2022-06-29
Bug 2099864: Reorder static resources to create RBAC first

jsafrane:
The privileged RBAC objects should be created first to make sure the CSI driver Deployment / DaemonSet can use privileged SCC.

cc @openshift/storage 

/hold
for testing


bertinatto (commented):
/lgtm

IMO we can reorder even if it doesn't completely fix the original issue.

jsafrane:
/hold cancel
no errors seen in 3 tests

PR #104, opened 2022-08-18
Bug 2108473: Warn if multiple credentials exists in secrets

gnufied:
There really shouldn't be more than one set of credentials in OCP secrets because CSI driver can't support more than one vcenter anyways.

For now lets warn the users, so as it is easier to debug. 

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2108473


bertinatto (commented):
/lgtm

PR #107, opened 2022-09-06
STOR-1050: Enable topology aware provisioning

gnufied:
Implement changes for topology aware dynamic provisioning.

Fixes https://issues.redhat.com/browse/STOR-1050


pkg/operator/vspherecontroller/driver_starter.go
```diff
@@ -115,6 +117,7 @@ func (c *VSphereController) createCSIDriver() {
 			trustedCAConfigMap,
 			c.apiClients.ConfigMapInformer,
 		),
+		WithConfigMapDaemonSetAnnotationHook("vsphere-csi-config", defaultNamespace, c.apiClients.ConfigMapInformer),
```
gnufied, line 120: Even though Daemonset spec does not have to be modified, it appears that drivers must be restarted for new `CSINode` values to be generated with updated topology values.

assets/controller.yaml
```diff
@@ -117,16 +117,15 @@ spec:
...
             # - --extra-create-metadata=true
             - --v=${LOG_LEVEL}
             - --leader-election
             - --leader-election-lease-duration=${LEADER_ELECTION_LEASE_DURATION}
             - --leader-election-renew-deadline=${LEADER_ELECTION_RENEW_DEADLINE}
             - --leader-election-retry-period=${LEADER_ELECTION_RETRY_PERIOD}
             - --http-endpoint=localhost:8202
-            # needed only for topology aware setup
-            # - "--strict-topology"
+            - "--strict-topology"
```
RomanBednar: Why is this uncommented here and then commented back in this [commit](https://github.com/openshift/vmware-vsphere-csi-driver-operator/pull/107/commits/7d4853dcc988b4222869aa81184113df4973d2ca)? 

Also if this should be uncommented it must not be quoted.

pkg/operator/driverfeaturescontroller.go
```diff
@@ -0,0 +1,128 @@
...
+		clusterCSIDriverInformer.Informer(),
+	).WithSync(
+		c.Sync,
+	).ResyncEvery(
+		time.Minute,
+	).WithSyncDegradedOnError(
+		operatorClient,
+	).ToController(
+		c.name,
+		recorder.WithComponentSuffix("feature-config-controlller-"+strings.ToLower(name)),
```
RomanBednar, line 69:
```suggestion
		recorder.WithComponentSuffix("feature-config-controller-"+strings.ToLower(name)),
```

assets/csi_cloud_config.ini
```diff
@@ -0,0 +1,7 @@
+[Global]
```
RomanBednar, line 1: Isn't assets directory intended only for OpenShift object definitions?

pkg/operator/vspherecontroller/driver_starter.go
```diff
@@ -127,6 +130,34 @@ func (c *VSphereController) createCSIDriver() {
...
+	if vsphereConfig.VSphere != nil {
+		topologyCategories := vsphereConfig.VSphere.TopologyCategories
+		if len(topologyCategories) > 0 {
+			containers := deployment.Spec.Template.Spec.Containers
+			for i := range containers {
+				if containers[i].Name != "csi-provisioner" {
+					continue
+				}
+
+				containers[i].Args = append(containers[i].Args, "--feature-gates=Topology=true", "--strict-topology")
```
RomanBednar, line 153: If I understand it correctly, the hook gets called in deployment controller and just modifies the manifest we provide if needed. So this hook can append `"--strict-topology"` flag to `"csi-provisioner"` container, but it will do so at runtime. 

So if the deployment manifest we provide is ever changed either by removing the flag, or uncommenting it, which would cause the flag to be present twice, it will break. Not sure we have that covered in CI.

assets/vsphere_cloud_config.yaml
```diff
@@ -4,9 +4,7 @@ metadata:
   namespace: openshift-cluster-csi-drivers
   name: vsphere-csi-config
 data:
-  cloud.conf: |
-    [Global]
-    cluster-id = "${CLUSTER_ID}"
-    [VirtualCenter "${VCENTER}"]
-    insecure-flag = "true"
-    datacenters = "${DATACENTERS}"
+  cloud.conf: ""
```
RomanBednar, line 7: This will get populated by what's in the ini file (after replacing the strings in it based on infa and vsphere config). Maybe using go templating here would make it more readable? Or maybe there's a good reason to use ini here?

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -28,41 +34,52 @@ const (
 	cloudConfigNamespace = "openshift-config"
 )
 
+var (
+	emptyDriverConfig = opv1.CSIDriverConfigSpec{}
```
jsafrane, line 38: I'm not a big fan of such global vars - IMO it does not save much `if`s below.

pkg/operator/driverfeaturescontroller.go
```diff
@@ -0,0 +1,128 @@
...
+		return err
+	}
+
+	defaultFeatureConfigMap := resourceread.ReadConfigMapV1OrDie(d.manifest)
+
+	driverConfig := clusterCSIDriver.Spec.DriverConfig
+	emptyDriverConfig := opv1.CSIDriverConfigSpec{}
+	if driverConfig != emptyDriverConfig {
+		vsphereConfig := driverConfig.VSphere
+		if vsphereConfig != nil && len(vsphereConfig.TopologyCategories) > 0 {
```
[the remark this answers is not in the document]
JoelSpeed: Driver config spec is a discriminated union, you need a switch here on the discriminant of the union rather than comparing to an empty config.

If you only care about one entry, then you don't need to, but normally what you'd expect is

```
switch driverConfig.DriverType{
case opv1.VSphereDriverType:
  // Do vSphere things here based on if the vSphere config was provided or not
default:
 // My controller /doesdoesn't understand this case, do the appropriate thing
}
```

go.mod
```diff
@@ -108,8 +109,12 @@ require (
...
 	sigs.k8s.io/json v0.0.0-20211208200746-9f7c6b3444d2 // indirect
 	sigs.k8s.io/kube-storage-version-migrator v0.0.4 // indirect
-	sigs.k8s.io/structured-merge-diff/v4 v4.2.1 // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 	sigs.k8s.io/yaml v1.2.0 // indirect
 )
 
 replace github.com/dgrijalva/jwt-go => github.com/golang-jwt/jwt v3.2.1+incompatible
+
+replace github.com/openshift/api => github.com/gnufied/api v0.0.0-20220907153857-5a99faf61076
```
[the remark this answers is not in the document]
gnufied: Yeah I do. I placed the PR on hold. Good catch!

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -149,3 +162,38 @@ func (c TargetConfigController) sync(ctx context.Context, syncContext factory.Sy
...
+	} {
+		csiConfigString = strings.ReplaceAll(csiConfigString, pattern, value)
+	}
+
+	csiConfig, err := iniv1.Load([]byte(csiConfigString))
+	if err != nil {
+		return nil, fmt.Errorf("error loading ini file: %v", err)
+	}
+
+	driverConfig := clusterCSIDriver.Spec.DriverConfig
```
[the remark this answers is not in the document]
gnufied: I don't know. I could try but yeah driver crashing with empty value of `topology-categories` was exactly what I was worried about and hence I chose to parse and load the ini file separately.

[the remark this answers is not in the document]
gnufied: Also - my motivation was to reduce amount of breakage that can happen from future changes. Who is to say, if driver handles empty topology now, it won't break in future. :-) So to be on safer side..

assets/controller.yaml
gnufied: oh sorry, those are intermediate commits. End result should be obviously be commented.

assets/csi_cloud_config.ini
gnufied: I don't think so, there is a rule like that.

pkg/operator/vspherecontroller/driver_starter.go
gnufied: fair point. Although - I think most argument parsers will still parse these and set values only once. I *could* check for exact string being already in the arguments, but I am afraid this is also error prone. What happens if these flags get renamed?

assets/vsphere_cloud_config.yaml
gnufied: I do not fully understand it. How will go templating help? See comment from @jsafrane below - https://github.com/openshift/vmware-vsphere-csi-driver-operator/pull/107#discussion_r965872534

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
gnufied: fixed.

pkg/operator/driverfeaturescontroller.go
```diff
@@ -0,0 +1,128 @@
...
+			existingData := defaultFeatureConfigMap.Data
```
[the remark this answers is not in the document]
gnufied: fixed

pkg/operator/driverfeaturescontroller.go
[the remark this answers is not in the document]
gnufied: fixed

assets/vsphere_cloud_config.yaml
RomanBednar: His suggestion is basically what I had in mind too, it does not have to be go template but it would look nice imho. It's just an idea to make it more readable, something similar to Helm templates but we pass in the values from code and not values.yaml. So the default value for each key could be `""` if it does not break anything and inject some real values only if that's appropriate - that's what @jsafrane meant I believe.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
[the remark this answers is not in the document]
RomanBednar: So the only concern is the use of empty values, and so it's more safe to have `cloud.conf: ""` by default? What if only some of the values in cloud conf are empty? Should it be an error or will it work fine? I see few potential issues in the current version:

1. ` applyClusterCSIDriverChange()` does not validate the values before doing `.ReplaceAll()` so it does not avoid empty strings (maybe it should have it regardless of ini vs template approach?)

2.  the resulting config map is obscured, it's very diffucult to tell what it will provide under `data:` and the label is hidden completely. If we go with a template instead we could use a control variable and conditions, something like this would be more transparent and robust:

```
apiVersion: v1
kind: ConfigMap
metadata:
  namespace: openshift-cluster-csi-drivers
  name: vsphere-csi-config
  {{ - if .Values.ValidConfig }}
  labels: 
    "topology-categories": {{ .Values.TopologyCategories }}
  {{ - end }}
data:
{{ - if .Values.ValidConfig }}
  [Global]
  cluster-id = {{ .Values.Id }}
  [VirtualCenter {{ .Values.VCenter }}]
  insecure-flag = "true"
  datacenters = {{ .Values.Datacenters }}
{{ else }}
  cloud.conf: ""
{{ - end }}
```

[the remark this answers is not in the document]
gnufied: >The resulting config map is obscured, it's very diffucult to tell what it will provide under data: and the label is hidden completely.

hmm, I am not sure I like the templatized YAML file. It is also slightly wrong btw. The `topology-categories` has to be in `data` field and part of ini content, not the configmap label. Anyways - do you think this is cleaner? I am not so sure, I have a long mistrust of templatizing YAMLs in a language that is not whitespace sensitive (like go). It is too easy to make a mistake and get indentation wrong.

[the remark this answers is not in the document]
RomanBednar: I see, I misinterpreted [this](https://github.com/openshift/vmware-vsphere-csi-driver-operator/pull/107/files#diff-0af15f182aea9898e784fc589175ef4c2411c9391bb5a1f41327ba7c3e2172f5R182) line. No need to go this route, it just seemed like we need to do a very similar thing to what Helm does. Surely it's prone to indentation errors, but it might be a good tradeoff for the advantages.

If we're quite certain this will not grow and more logic will be needed around generating config maps (or other resources) it's probably fine to go with ini and string replacement.

pkg/operator/driverfeaturescontroller.go
```diff
@@ -0,0 +1,124 @@
...
+		clusterCSIDriverInformer.Informer(),
+	).WithSync(
+		c.Sync,
+	).ResyncEvery(
+		time.Minute,
+	).WithSyncDegradedOnError(
+		operatorClient,
+	).ToController(
+		c.name,
+		recorder.WithComponentSuffix("feature-config-controlller-"+strings.ToLower(name)),
```
jsafrane, line 69: typo: `controlller`

pkg/operator/driverfeaturescontroller.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+	availableCondition := opv1.OperatorCondition{
+		Type:   d.name + opv1.OperatorStatusTypeAvailable,
+		Status: opv1.ConditionTrue,
+	}
+
+	progressingCondition := opv1.OperatorCondition{
+		Type:   d.name + opv1.OperatorStatusTypeProgressing,
+		Status: opv1.ConditionFalse,
+	}
```
jsafrane, line 115: I think it's useless to set Progressing condition if it's never True. And IMO Available is quite useless too, this controlled could be fine just with Degraded, if anything goes wrong.
StaticResourceController does not have Available nor Progressing either.

pkg/operator/starter.go
```diff
@@ -88,25 +95,54 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
...
 		cloudConfigBytes,
+		csiConfigBytes,
 		kubeClient,
 		kubeInformersForNamespaces,
 		operatorClient,
 		configInformers,
+		clusterCSIDriverInformer,
+		controllerConfig.EventRecorder,
+	)
+	featureConfigBytes, err := assets.ReadFile("vsphere_features_config.yaml")
```
jsafrane, line 116: nit: add an empty line + delete the one below.

pkg/operator/utils/topology.go
```diff
@@ -0,0 +1,22 @@
...
+func GetTopologyCategories(clusterCSIDriver *opv1.ClusterCSIDriver) []string {
+	driverConfig := clusterCSIDriver.Spec.DriverConfig
+	if driverConfig.DriverType == opv1.VSphereDriverType {
+		vSphereConfig := driverConfig.VSphere
+		if vSphereConfig != nil && len(vSphereConfig.TopologyCategories) > 0 {
+			topologyCategories := vSphereConfig.TopologyCategories
+			// sort the topologyCategories so as they are stable and don't change their order
+			sort.Slice(topologyCategories, func(i, j int) bool {
+				return topologyCategories[i] < topologyCategories[j]
+			})
```
jsafrane, line 17: Why is this sorted? Should `topology-categories` in the ini file have the same order of labels as this `TopologyCategories` field? IMO it would be quite confusing if the operator changes the order.

pkg/operator/vspherecontroller/vspherecontroller.go
```diff
@@ -65,6 +65,7 @@ const (
 	resyncDuration                    = 20 * time.Minute
 	envVMWareVsphereDriverSyncerImage = "VMWARE_VSPHERE_SYNCER_IMAGE"
 	storageClassControllerName        = "VMwareVSphereDriverStorageClassController"
+	driverName                        = "csi.vsphere.vmware.com"
```
jsafrane, line 68: This can be removed and `utils.VSphereDriverName` can be used instead.

assets/vsphere_cloud_config.yaml
jsafrane, line 7: Please add a comment that cloud.conf is loaded from csi_cloud_config.ini from the operator

assets/csi_cloud_config.ini
```diff
@@ -0,0 +1,7 @@
...
+cluster-id = "${CLUSTER_ID}"
+
+[VirtualCenter "${VCENTER}"]
+insecure-flag = "true"
+datacenters = "${DATACENTERS}"
+
```
jsafrane, line 7: Please add a comment that [Labels] is added by the operator

pkg/operator/driverfeaturescontroller.go
```diff
@@ -0,0 +1,124 @@
...
+	if err != nil {
+		return err
+	}
+
+	defaultFeatureConfigMap := resourceread.ReadConfigMapV1OrDie(d.manifest)
+
+	topologyCategories := utils.GetTopologyCategories(clusterCSIDriver)
+	if len(topologyCategories) > 0 {
+		requiredData := defaultFeatureConfigMap.Data
+		requiredData["improved-volume-topology"] = "true"
```
jsafrane, line 99: Please add a comment to `assets/vsphere_features_config.yaml` that `improved-volume-topology` is set by the operator. And maybe add explicit `improved-volume-topology: "false"` there.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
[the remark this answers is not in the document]
jsafrane: I am not a big fan of huge templates either. IMO it's enough if we add comments to assets that certain fields are set by the operator.

pkg/operator/utils/topology.go
gnufied: The topology categories in API is an array whereas one in ini file is a single value string. It is possible that whenever we convert the array to the string, order may change and it could result in driver restart. That is why I am sorting them, so as to keep the order "stable".

From driver perspective, I don't think order in which categories are specified matters (I tested by changing the order and it works the same).

gnufied:
@jsafrane updated PTAL

assets/vsphere_cloud_config.yaml
gnufied: fixed

pkg/operator/driverfeaturescontroller.go
gnufied: fixed

pkg/operator/utils/topology.go
jsafrane: > It is possible that whenever we convert the array to the string, order may change

How is it possible? It's just an array / slice and go does not reshuffle them randomly.

gnufied: Hmm, that is fair. I have removed sorting for now.

duanwei33:
Following the [steps](https://docs.google.com/document/d/16qG2dwdBDJ8BKxUhP0QERgUqgPbVDU7L8bW2D2WwZjE/edit#heading=h.oygf33drg958)  to verified in single zone cluster passed: 

Checking csinode with topology:
```
$ oc get csinode ocp-storage-8mgh7-worker-d5jqb -ojson | jq .spec.drivers[].topologyKeys
[
  "topology.csi.vmware.com/openshift-region",
  "topology.csi.vmware.com/openshift-zone"
]
```
Checking pv with topology:
```
$ oc get pv pvc-df1c6674-8e16-4bfe-ae77-31642b0bcb2c -ojson | jq .spec.nodeAffinity
{
  "required": {
    "nodeSelectorTerms": [
      {
        "matchExpressions": [
          {
            "key": "topology.csi.vmware.com/openshift-region",
            "operator": "In",
            "values": [
              "openshift-region-a"
            ]
          },
          {
            "key": "topology.csi.vmware.com/openshift-zone",
            "operator": "In",
            "values": [
              "openshift-zone-a"
            ]
          }
        ]
      }
    ]
  }
}
```

jsafrane:
@duanwei33, can you please mark this as qa-approved?

PR #108, opened 2022-09-09
OCPBUGS-1067: Recreate storageclass in case it was deleted

gnufied:


pkg/operator/vspherecontroller/vspherecontroller.go
```diff
@@ -193,6 +193,17 @@ func (c *VSphereController) sync(ctx context.Context, syncContext factory.SyncCo
...
+	// if we created storageclass before then we should keep resyncing the storageclass in case it gets deleted.
+	if c.storageClassController != nil && c.storageClassController.StorageClassCreated {
+		storageClassAPIDeps := c.getCheckAPIDependency(infra)
+		err = c.storageClassController.Sync(ctx, c.vSphereConnection, storageClassAPIDeps)
+		// storageclass sync will only return error if somehow updating conditions fails, in which case
+		// we can return error here and degrade the cluster
+		if err != nil {
+			return err
+		}
+	}
```
dobsonj, line 205: Is it possible to get rid of this second call to sync the storageclass with a bit of refactoring?
https://github.com/openshift/vmware-vsphere-csi-driver-operator/blob/fcca127a6900b9709bf88fe6e1ef7b99d48ba6e1/pkg/operator/vspherecontroller/vspherecontroller.go#L233-L242
Right now we'll end up syncing the SC twice in the same loop in some cases. Probably not a big deal, and I could live with it.
But would it work to instead have a boolean like `ShouldSyncSC` that only gets set to true after all cluster checks succeed, and then check that boolean here, and have just this one call to `c.storageClassController.Sync`?

dobsonj: I guess the downside of that variation is that we might not create the SC immediately after the cluster checks run the first time?

gnufied: Okay so I thought long and hard about this and sort of have fixed it. And the new idea is - only sync storageclasses if we are going to install CSI driver (or if we have previously installed CSI driver). I think this addresses previous issue where we were creating SCs even if environment did not support the CSI driver and at the same time, if someone deletes the SC, it should get recreated immediately.

dobsonj: Nice, I like this approach, and the code looks good. Just waiting for e2e tests to finish.

PR #110, opened 2022-09-23
OCPBUGS-1592: Add HTTP proxy to syncer container

jsafrane:
`vsphere-syncer` container needs HTTP proxy too.

cc @openshift/storage


bertinatto (commented):
/lgtm

PR #112, opened 2022-09-28
OCPBUGS-1805: multi-zone installs may not provision storage

rvanderp3:
depending on which datacenter storage is being provisioned, the CSI driver may be checking if the VM is in the wrong datacenter.

### DIFF
--- a/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
+++ b/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
@@ -3,13 +3,15 @@ package targetconfigcontroller
 import (
 	"context"
 	"fmt"
+	"strings"
+	"time"
+
 	ocpconfigv1 "github.com/openshift/api/config/v1"
 	clustercsidriverinformer "github.com/openshift/client-go/operator/informers/externalversions/operator/v1"
 	clustercsidriverlister "github.com/openshift/client-go/operator/listers/operator/v1"
 	"github.com/openshift/vmware-vsphere-csi-driver-operator/pkg/operator/utils"
 	corev1 "k8s.io/api/core/v1"
-	"strings"
-	"time"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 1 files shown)


pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -115,10 +116,19 @@ func (c TargetConfigController) sync(ctx context.Context, syncContext factory.Sy
...
 
 	cmString := string(c.manifest)
+
+	virtualCenterIPs := sets.StringKeySet(cfg.VirtualCenter)
+
+	if len(virtualCenterIPs) > 1 || len(virtualCenterIPs) == 0 {
+		return fmt.Errorf("cloud config %s/%s must define a single VirtualCenter", cloudConfigNamespace, cloudConfig.Name)
+	}
+
+	virtualCenterIP := virtualCenterIPs.List()[0]
```
dobsonj, line 126: @gnufied @bertinatto Do you see a way for us to actually support multiple datacenters? Or is this the best option we have for now (throw a clear error)?

rvanderp3: The restriction here is actually a single vCenter.  With multi-zone support, it is very possible a customer may have multiple datacenters.

rvanderp3: and that could probably just be refactored to `if len(virtualCenterIPs) != 1` . :facepalm:

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -133,6 +134,25 @@ func (c TargetConfigController) sync(ctx context.Context, syncContext factory.Sy
...
+
+	for pattern, value := range map[string]string{
+		"${CLUSTER_ID}":  infra.Status.InfrastructureName,
+		"${VCENTER}":     virtualCenterIP,
+		"${DATACENTERS}": cfg.VirtualCenter[virtualCenterIP].Datacenters,
+	} {
+		cmString = strings.ReplaceAll(cmString, pattern, value)
+	}
+
+	requiredCM := resourceread.ReadConfigMapV1OrDie([]byte(cmString))
```
gnufied, line 155: This is making change in outdated code. Can you check code again and update?

rvanderp3: hi @gnufied odd, i just rebased

rvanderp3: i'll check again

rvanderp3:
/hold

placing a hold while performing additional multi-zone testing

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
gnufied: So we traditionally picked this from workspace and now we are changing it to vcenter section. Could we break anyone? Such as UPI installs or people who have modified their cloud-configs? Shouldn't workspace datacenter and `VirtualCenter` datacenter be the same?

PR #115, opened 2022-10-18
Add multiple datacenters support

gnufied:


pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -163,10 +164,29 @@ func (c TargetConfigController) applyClusterCSIDriverChange(
...
+	if len(virtualCenterIPs) != 1 {
+		return nil, fmt.Errorf("cloud config must define a single VirtualCenter")
+	}
+
+	var vCenterDatacenter string
+
+	virtualCenterIP := virtualCenterIPs.List()[0]
+	if virtualCenterConfig, ok := sourceCFG.VirtualCenter[virtualCenterIP]; ok {
+		vCenterDatacenter = virtualCenterConfig.Datacenters
+	}
```
bertinatto, line 180: This could be simplified to:
```
	var vCenterDatacenters string
	for _, v := range sourceCFG.VirtualCenter {
		vCenterDatacenters = v.Datacenters
		break // Optional, we know this look iterates only once because of the check above.
	}
```

pkg/operator/utils/topology.go
```diff
@@ -14,3 +19,19 @@ func GetTopologyCategories(clusterCSIDriver *opv1.ClusterCSIDriver) []string {
...
+
+	if len(virtualCenterIPs) != 1 {
+		return nil, fmt.Errorf("cloud config must define a single VirtualCenter")
+	}
+
+	virtualCenterIP := virtualCenterIPs.List()[0]
+	if virtualCenterConfig, ok := config.VirtualCenter[virtualCenterIP]; ok {
+		datacenters = strings.Split(virtualCenterConfig.Datacenters, ",")
+	}
+	return datacenters, nil
```
bertinatto, line 36: Can we unify with the common logic from `targetconfigcontroller.go`?

assets/vsphere_features_config.yaml
```diff
@@ -5,6 +5,7 @@ data:
   "online-volume-extend": "true"
   "block-volume-snapshot": "true"
   "improved-volume-topology": "false" # operator will set this value to true if cluster supports CSI topology
+  "use-csinode-id": "true"
```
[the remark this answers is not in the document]
gnufied: I don't think we need this change yet, but it is being used as a default in upstream deployments, so I turned it on. I think the idea is, not to depend on `providerID` field in node object, which requires intree or external CCM. So enabling `use-csinode-id` seems to remove that requirement.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
gnufied: hmm, seems similar enough. not sure if simpler.

pkg/operator/utils/topology.go
gnufied: fixed

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -163,10 +164,29 @@ func (c TargetConfigController) applyClusterCSIDriverChange(
 	infra *ocpconfigv1.Infrastructure, sourceCFG vsphere.VSphereConfig, clusterCSIDriver *opv1.ClusterCSIDriver) (*corev1.ConfigMap, error) {
 	csiConfigString := string(c.csiConfigManifest)
 
+	datacenters := sourceCFG.Workspace.Datacenter
+
+	virtualCenterIPs := sets.StringKeySet(sourceCFG.VirtualCenter)
+
+	if len(virtualCenterIPs) != 1 {
+		return nil, fmt.Errorf("cloud config must define a single VirtualCenter")
```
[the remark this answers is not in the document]
gnufied: fixed

pkg/operator/vspherecontroller/checks/check_nodes.go
```diff
@@ -236,33 +237,40 @@ func getVM(ctx context.Context, checkOpts CheckArgs, node *v1.Node) (*mo.Virtual
...
-		return nil, fmt.Errorf("failed to access Datacenter %s: %s", vmConfig.Workspace.Datacenter, err)
+		return nil, err
 	}
 
-	// Find VM reference in the datastore, by UUID
-	s := object.NewSearchIndex(dc.Client())
 	vmUUID := strings.ToLower(strings.TrimSpace(strings.TrimPrefix(node.Spec.ProviderID, "vsphere://")))
+	for _, dcName := range dataCenterNames {
+		// Find datastore
+		finder := find.NewFinder(vmClient, false)
```
bertinatto, line 248: Can this be moved outside the loop?

pkg/operator/vspherecontroller/checks/check_nodes.go
```diff
@@ -236,33 +237,40 @@ func getVM(ctx context.Context, checkOpts CheckArgs, node *v1.Node) (*mo.Virtual
...
+		// Find VM reference in the datastore, by UUID
+		s := object.NewSearchIndex(dc.Client())
+
+		svm, err := s.FindByUuid(ctx, dc, vmUUID, true, nil)
+		if err != nil {
+			return nil, fmt.Errorf("failed to find VM %s by UUID %s: %s", node.Name, vmUUID, err)
+		}
+		if svm == nil {
+			continue
+		}
```
bertinatto, line 263: When the VM is not found, will `err == nil` and `svm == nil`?  Won't it return some sort of "not found" error here?

gnufied: Yes that is correct. It doesn't seem to return an error when vm is not found. I have tested it.

bertinatto (commented):
/lgtm

duanwei33:
/label qe-approved 
Verified in multiple datacenters support and talked to Hemant offline, sorry I missed the labeling.

jsafrane:
/retest

`ProvisioningFailed: failed to provision volume with StorageClass "e2e-fsgroupchangepolicy-1376-e2e-sclcdnb": rpc error: code = DeadlineExceeded desc = context deadline exceeded` ???

gnufied:
The only failing tests were because of generic ephemeral failures.

PR #118, opened 2022-11-04
[release-4.12] OCPBUGS-3281: set TLS cipher suites in Kube RBAC sidecars

dobsonj:
/hold for https://github.com/openshift/library-go/pull/1414
I need to bump library-go once that merges.


dobsonj:
/unhold
library-go PR merged, and I added a new commit here to get the new version.
/cc @openshift/storage

PR #121, opened 2022-12-06
STOR-1049: Add topology aware sc

gnufied:
Fixes https://issues.redhat.com/browse/STOR-1049


gnufied:
cc @rvanderp3 @jcpowermac

pkg/operator/storageclasscontroller/vmware.go
```diff
@@ -89,6 +89,59 @@ func (v *storagePolicyAPI) getDefaultDatastore(ctx context.Context) (*mo.Datasto
...
+	if err != nil {
+		return nil, fmt.Errorf("error getting properties of datastore %s: %v", dsName, err)
+	}
+	return &dsMo, nil
+}
+
+func (v *storagePolicyAPI) createZonalStoragePolicy(ctx context.Context) (string, error) {
+	var err error
+	failureDomains := v.infra.Spec.PlatformSpec.VSphere.FailureDomains
+	if len(failureDomains) <= 1 {
```
jcpowermac, line 121: Why would one failure domain be an error?

gnufied: Not a hard error as such, but given that this gets called only when `len(failureDomain) > 1`, it would be strange to have it less than one now.

pkg/operator/storageclasscontroller/vmware.go
```diff
@@ -89,6 +89,59 @@ func (v *storagePolicyAPI) getDefaultDatastore(ctx context.Context) (*mo.Datasto
...
+func (v *storagePolicyAPI) getDatastore(ctx context.Context, dcName string, dsName string) (*mo.Datastore, error) {
+	vmClient := v.vcenterApiConnection.Client
+	finder := find.NewFinder(vmClient.Client, false)
+
+	dc, err := finder.Datacenter(ctx, dcName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to access datacenter %s: %s", dcName, err)
+	}
+
+	finder = find.NewFinder(vmClient.Client, false)
```
rvanderp3, line 101: do we need a new finder here?

pkg/operator/storageclasscontroller/vmware.go
bertinatto: nit: the function name says it creates a zonal storage policy (singular), but the code creates multiple storage policies.

pkg/operator/storageclasscontroller/vmware.go
```diff
@@ -89,6 +89,58 @@ func (v *storagePolicyAPI) getDefaultDatastore(ctx context.Context) (*mo.Datasto
...
+		dataCenter := failureDomain.Topology.Datacenter
+		dataStore := failureDomain.Topology.Datastore
+		ds, err := v.getDatastore(ctx, dataCenter, dataStore)
+		if err != nil {
+			return v.policyName, fmt.Errorf("unable to fetch datastore %s: %v", dataStore, err)
+		}
+		err = v.createOrUpdateTag(ctx, ds)
+		if err != nil {
+			return v.policyName, fmt.Errorf("error creating or updating tag %s: %v", v.tagName, err)
+		}
```
bertinatto, line 134: Instead of stopping the execution, should it try the next one?

pkg/operator/storageclasscontroller/vmware.go
gnufied: No the code creates only one storage policy. It creates and attaches tags to multiple datastores found in each failure-domain.

bertinatto: Sorry, I missed that

bertinatto (commented):
Just a comment related to error aggregation, otherwise LGTM (not tagging to give @jsafrane a chance to review).

pkg/operator/storageclasscontroller/vmware.go
gnufied: I think it might be better to hard error out, rather than continuing with tagging whatever datastores we could access, because that could result in a storagepolicy which has unpredictable behaviour.

bertinatto: My point was when you have say 5 failureDomains/datastores, then you successfully tag 3 of them and the 4th fail. Sounds like it's better to either:
1. Try tagging all of them and return an aggregated error
2. Return early in case of error, but don't leave behind tagged datastores

pkg/operator/storageclasscontroller/vmware.go
```diff
@@ -100,6 +152,11 @@ func (v *storagePolicyAPI) createStoragePolicy(ctx context.Context) (string, err
 		return v.policyName, nil
 	}
 
+	vSphereInfraConfig := v.infra.Spec.PlatformSpec.VSphere
+	if vSphereInfraConfig != nil && len(vSphereInfraConfig.FailureDomains) > 1 {
+		return v.createZonalStoragePolicy(ctx)
+	}
+
```
jsafrane, line 159: `createZonalStoragePolicy` and rest of `createStoragePolicy` below are almost the same, `createZonalStoragePolicy` only loops / tags over more datastores. Would it be possible to join them together somehow? Like `getZonalDatastores()` + `getDefaultDatastore()` and then a common tagging loop + `createStorageProfile` over the result.

pkg/operator/utils/topology.go
```diff
@@ -4,12 +4,26 @@ import (
...
+	defaultOpenshiftZoneCategory   = "openshift-zone"
+	defaultOpenshiftRegionCategory = "openshift-region"
+)
+
+func GetTopologyCategories(clusterCSIDriver *opv1.ClusterCSIDriver, infra *v1.Infrastructure) []string {
+	vSpherePlatformConfig := infra.Spec.PlatformSpec.VSphere
+	if vSpherePlatformConfig != nil {
+		failureDomains := vSpherePlatformConfig.FailureDomains
+		if len(failureDomains) > 0 {
+			return []string{defaultOpenshiftZoneCategory, defaultOpenshiftRegionCategory}
```
jsafrane, line 23: I'd prefer a warning when both infra and clusterCSIDriver specify topology. Maybe with a metric + alert, but that probably does not belong to this function.

pkg/operator/storageclasscontroller/vmware.go
gnufied: okay I have aggregated the errors and returned them. I think that is a good idea in case we have multiple datastores and we don't have access to them, so as rather than returning one error at a time - we will return error about all inaccessible datastores at the same time. 

I have pushed that change to - https://github.com/openshift/vmware-vsphere-csi-driver-operator/pull/125 however, which also includes #121

pkg/operator/storageclasscontroller/vmware.go
gnufied: okay I have unified that code. But I have pushed my new changes to https://github.com/openshift/vmware-vsphere-csi-driver-operator/pull/125/files#diff-cd720ec01eefb8566ee378bc53aef398da5550c725aeacd65aba7efb2b39e311R146 

That branch already includes commit from this branch and has more tests and ensures tags are recreated if deleted etc.

gnufied:
I am going to close this favour of https://github.com/openshift/vmware-vsphere-csi-driver-operator/pull/125

/close

pkg/operator/utils/topology.go
gnufied: https://issues.redhat.com/browse/STOR-1123

PR #125, opened 2022-12-08
STOR-1121:Make policy creation idempotent

gnufied:
This PR makes policy creation idempotent and allows operator to discover new failure-domains and tag datastore on them, if necessary.

Also adds tests.

Includes commit for: https://github.com/openshift/vmware-vsphere-csi-driver-operator/pull/121

Here is a summary of changes:
- Add vcsim based unit tests for tag, category and policy creation.
- Fix tags if accidently removed from datastore or category gets deleted.
- Allow tagging of new failure-domains if infra object is  edited and new failure-domains are added.
- In general make the code more failure-proof (I think. :D )

Fixes https://issues.redhat.com/browse/STOR-1121

### DIFF
--- a/pkg/operator/storageclasscontroller/vmware.go
+++ b/pkg/operator/storageclasscontroller/vmware.go
@@ -132,9 +132,6 @@ func (v *storagePolicyAPI) getDatastore(ctx context.Context, dcName string, dsNa
 func (v *storagePolicyAPI) createZonalStoragePolicy(ctx context.Context) (string, error) {
 	var err error
 	failureDomains := v.infra.Spec.PlatformSpec.VSphere.FailureDomains
-	if len(failureDomains) <= 1 {
-		return "", fmt.Errorf("cluster has only one failure domain defined")
-	}
 
 	var aggregatedErrors []error
 
@@ -193,7 +190,7 @@ func (v *storagePolicyAPI) createStoragePolicy(ctx context.Context) (string, err
 	}
 
 	vSphereInfraConfig := v.infra.Spec.PlatformSpec.VSphere
-	if vSphereInfraConfig != nil && len(vSphereInfraConfig.FailureDomains) > 1 {
+	if vSphereInfraConfig != nil && len(vSphereInfraConfig.FailureDomains) > 0 {
 		return v.createZonalStoragePolicy(ctx)
 	}
 
--- a/pkg/operator/storageclasscontroller/vmware_test.go
+++ b/pkg/operator/storageclasscontroller/vmware_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"testing"
 
+	v1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/vmware-vsphere-csi-driver-operator/pkg/operator/testlib"
 	"github.com/openshift/vmware-vsphere-csi-driver-operator/pkg/operator/vclib"
 )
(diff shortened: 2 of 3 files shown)


pkg/operator/storageclasscontroller/vmware.go
```diff
@@ -205,6 +328,7 @@ func (v *storagePolicyAPI) createStorageProfile(ctx context.Context) error {
 
 	var policySpec types.PbmCapabilityProfileCreateSpec
 	policySpec.Name = v.policyName
+	policySpec.Category = "REQUIREMENT"
```
jsafrane, line 331: This change is surprising. Was the storage profile created in OCP 4.12 correct without this field?

pkg/operator/storageclasscontroller/vmware.go
```diff
@@ -89,33 +104,135 @@ func (v *storagePolicyAPI) getDefaultDatastore(ctx context.Context) (*mo.Datasto
...
+	}
+
+	// skip creation of tags on a datastore if it already has been tagged.
+	if v.policyCreated && v.checkForTagOnDatastore(ctx, ds) {
+		return nil
+	}
+
+	err = v.createOrUpdateTag(ctx, ds)
+	if err != nil {
+		return fmt.Errorf("error creating or updating tag %s: %v", v.tagName, err)
```
jsafrane, line 180: is the datastore name already part of `err` from `createOrUpdateTag` / `tagManager.AttachTag` inside? If not, it could be useful for a cluster admin in a multi-datastore cluster.

pkg/operator/storageclasscontroller/vmware.go
```diff
@@ -89,33 +104,135 @@ func (v *storagePolicyAPI) getDefaultDatastore(ctx context.Context) (*mo.Datasto
...
+}
+
+func (v *storagePolicyAPI) attachTags(ctx context.Context, dcName, dsName string) error {
+	ds, err := v.getDatastore(ctx, dcName, dsName)
+	if err != nil {
+		return fmt.Errorf("unable to fetch datastore %s: %v", dsName, err)
+	}
+
+	// skip creation of tags on a datastore if it already has been tagged.
+	if v.policyCreated && v.checkForTagOnDatastore(ctx, ds) {
```
jsafrane, line 174: Why is `policyCreated` checked here?

pkg/operator/storageclasscontroller/vmware.go
```diff
@@ -89,33 +104,135 @@ func (v *storagePolicyAPI) getDefaultDatastore(ctx context.Context) (*mo.Datasto
...
 	}
 
 	if found {
-		klog.V(3).Infof("found existing storage policy %s", v.policyName)
-		return v.policyName, nil
+		v.policyCreated = true
+	}
+
+	vSphereInfraConfig := v.infra.Spec.PlatformSpec.VSphere
+	if vSphereInfraConfig != nil && len(vSphereInfraConfig.FailureDomains) > 1 {
```
jsafrane, line 196: Why `len(vSphereInfraConfig.FailureDomains) > 1` here? Is it really safe to assume that if there is just 1 `FailureDomains`, then it's equal to `DefaultDatastore` that is tagged below?

pkg/operator/storageclasscontroller/vmware.go
```diff
@@ -287,13 +449,23 @@ func updateAssociatedTypes(associatedTypes []string) []string {
...
+		if strings.HasPrefix(associableType, vim25Prefix) {
+			appendedTypes = append(appendedTypes, associableType)
+		} else {
+			appendedTypes = append(appendedTypes, vim25Prefix+associableType)
+		}
 	}
+	// We are sorting this array because some part of vSphere that applies these tags
+	// is confused by ordering and considers same associated types in different order as
+	// different associated type.
+	sort.Strings(appendedTypes)
```
jsafrane, line 469: Do we need to backport it to 4.12?

pkg/operator/storageclasscontroller/vmware_test.go
```diff
@@ -0,0 +1,190 @@
...
+	validateApiCallCount(t, storagePolicyAPIClient, map[string]int{
+		create_tag_api:      1,
+		create_category_api: 1,
+		update_category_api: 1,
+		attach_tag_api:      2,
+		create_profile_api:  1,
+	})
+}
+
+func validateApiCallCount(t *testing.T, vmwareAPI *storagePolicyAPI, expectedMap map[string]int) {
```
jsafrane, line 183: nit: validateAPICallCount

pkg/operator/storageclasscontroller/vmware.go
gnufied: I had to make this change mostly to keep `vcsim` client happy. It appears that - with real cluster, this field is automatically set and it works as expected. But with `vcsim` client - when we don't set this field, it caches the object "as it is" and then when we query for this profile - it can't find it. Because it expectes `Category == REQUIREMENT`. 

Not sure how to do get out of this. I could not modify cached profiles because `vcsim` package stores that as a private unexported field and  I can't do modify it...

gnufied: I also tested, it doesn't break real API calls - so I thought it is fine.

jsafrane: ack

pkg/operator/storageclasscontroller/vmware.go
gnufied: To be extra safe.. :-) I am happy to drop them.

pkg/operator/storageclasscontroller/vmware.go
gnufied: That is a good question. @rvanderp3 @jcpowermac - if a cluster has only one failure-domain, does that mean that its topology will be same as default topology? (i.e `datacenter` and `defaultdatastore` defined above?):

pkg/operator/storageclasscontroller/vmware.go
gnufied: I tested this some more and it appears that, this affects `vcsim` code more than real stuff. But still I would like to keep this change.

pkg/operator/storageclasscontroller/vmware.go

pkg/operator/storageclasscontroller/vmware.go
rvanderp3: not necessarily, the topology should define its applicable datastore.  the installer ensures that the topology is filled out with a datastore.  if they dont define a datastore for the topology, the [defaultDatastore is inserted there](https://github.com/openshift/installer/blob/master/pkg/types/vsphere/validation/platform.go#L82).  we are moving in the direction of all vsphere installs being zonal to reduce the maintenance and configuration complexity.

gnufied: fixed. I changed the check to check for `>0`

gnufied:
Test appears to be failing on - https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_vmware-vsphere-csi-driver-operator/125/pull-ci-openshift-vmware-vsphere-csi-driver-operator-master-e2e-vsphere/1612505020652064768

PR #129, opened 2023-01-11
STOR-1078: Add hostPaths necessary for SELinux mounts

jsafrane:
To support "mount -o context=XYZ", /etc/selinux and /sys/fs/selinux from the host must be present in the CSI driver container.

cc @openshift/storage


bertinatto (commented):
/lgtm

jsafrane:
/label px-approved
/label docs-approved
/label qe-approved
This is not an user-facing change + I got approval from QE for such a small change.

PR #130, opened 2023-01-12
OCPBUGS-5790: Allow cluster to be upgraded even if using multiple datacenters

gnufied:
Fixes https://issues.redhat.com/browse/OCPBUGS-5790

This is basically a manual cherry-pick of https://github.com/openshift/vmware-vsphere-csi-driver-operator/pull/123 because automatic cherry-pick was throwing tons of conflicts.

### DIFF
--- a/pkg/operator/utils/topology.go
+++ b/pkg/operator/utils/topology.go
@@ -1,0 +1,25 @@
+package utils
+
+import (
+	"fmt"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/legacy-cloud-providers/vsphere"
+)
+
+func GetDatacenters(config *vsphere.VSphereConfig) ([]string, error) {
+	datacenters := []string{config.Workspace.Datacenter}
+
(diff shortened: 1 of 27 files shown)


pkg/operator/utils/topology.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/legacy-cloud-providers/vsphere"
+)
+
+func GetDatacenters(config *vsphere.VSphereConfig) ([]string, error) {
```
dobsonj, line 11: Don't you need the changes in pkg/operator/vspherecontroller/checks/check_nodes.go too?
https://github.com/openshift/vmware-vsphere-csi-driver-operator/pull/123/files#diff-872043e7b350e043c36bbae56b122ac24f16df367350b832376f0e3a2a23582eL239-R240

gnufied: Good catch! Will fix.

gnufied: fixed.

PR #131, opened 2023-01-16
Bug 2106736: Add multiplePVsSameID capability

jsafrane:
This CSI driver can handle multiple PVs that have the same VolumeHandle used on the same node.

cc @openshift/storage


bertinatto (commented):
/lgtm

PR #138, opened 2023-02-07
OCPBUGS-5765: Populate migration datastore url

gnufied:
Populate migration datastore url to fix bug with random datastore being selected.

https://issues.redhat.com/browse/OCPBUGS-5765


jsafrane:
Please use `UPSTREAM: 2170: ` format for the commit messages, so we know where it comes from during the next rebase.

/jira refresh

gnufied:
> Please use UPSTREAM: 2170:  format for the commit messages, so we know where it comes from during the next rebase.

@jsafrane why? These changes are being made to operator and not the driver.

jsafrane:
sorry, I missed this is for the operator

/lgmt

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -102,46 +87,12 @@ func (c TargetConfigController) sync(ctx context.Context, syncContext factory.Sy
...
-	if err != nil {
-		return err
-	}
-
-	// TODO: check if configMap has been deployed and set appropriate conditions
-	_, _, err = resourceapply.ApplyConfigMap(ctx, c.kubeClient.CoreV1(), syncContext.Recorder(), requiredCM)
-	if err != nil {
-		return err
+	// if configmap is not created simply return
+	if !configMapState {
```
gnufied, line 94: @jsafrane @bertinatto please check this. This will cause targetconfig operator to only set conditions once configmaps is created. I have been debating, if we should remove this controller entirely. 

Could we miss a case where cluster was upgraded but has left over conditions?

jsafrane: There is [RemoveStaleConditionsController](https://github.com/openshift/library-go/blob/a467dafcef6988afbcce43400c339d7ca68a3298/pkg/operator/staleconditions/remove_stale_conditions.go#L19) that can remove orphaned conditions

pkg/operator/utils/shared_state.go
```diff
@@ -0,0 +1,47 @@
...
+	g.Lock()
+	defer g.Unlock()
+	g.csiConfigMapCreated = state
+}
+
+func (g *OperatorSharedState) SetDefaultDatastore(datastorURL string) {
+	g.Lock()
+	defer g.Unlock()
+	g.defaultDatastore = datastorURL
+}
```
bertinatto, line 47: I'm not sure if I understand why we need a global state. Typically a ConfigMap is used for that, but IIUC you're replacing it with a global object.

IMO it's a bit more complicated this way. Perhaps it's also error-prone, since the same information is stored in 2 different places?

assets/csi_cloud_config.ini
```diff
@@ -6,4 +6,5 @@ cluster-id = "${CLUSTER_ID}"
 [VirtualCenter "${VCENTER}"]
 insecure-flag = "true"
 datacenters = "${DATACENTERS}"
+migration-datastore-url = "${MIGRATION_DATASTORE_URL}"
```
bertinatto, line 9: Where will this datastore will be used?

gnufied: This datastore will be used when no datastore is set in intree SC and rather than current behaviour of randomly picking a datastore from all available datastores, the datastore specified here will be used.

pkg/operator/utils/shared_state.go
gnufied: Removed this particular field. The shared state is currently used for signalling to target_config controller that, the configmap has been created, so as it can remove stale conditions.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
gnufied: Added a variant of it, so as stale conditions can be removed.

jsafrane: This controller won't do anything until a config map is loaded, it will not even update the old conditions. How useful are these conditions then? Can we remove them right away?

gnufied: I have updated the code to remove older conditions right away.

jsafrane: I should have made myself more obvious, can you please remove the whole `TargetConfigController` and just run [RemoveStaleConditionsController](https://github.com/openshift/library-go/blob/a467dafcef6988afbcce43400c339d7ca68a3298/pkg/operator/staleconditions/remove_stale_conditions.go#L19) instead? Less code, more re-use.

gnufied: The code in both the controllers are not the same. I do not think, it makes sense to remove conditions every 1 minute.

PR #140, opened 2023-02-15
OCPBUGS-7482: Do not enable topology features if only one failure-domains is defined

gnufied:
Fixes https://issues.redhat.com/browse/OCPBUGS-7482


bertinatto (commented):
/lgtm

PR #141, opened 2023-02-22
OCPBUGS-7899: Bump sidecar timeouts for vsphere

gnufied:
Fixes https://issues.redhat.com/browse/OCPBUGS-7899

### DIFF
--- a/assets/controller.yaml
+++ b/assets/controller.yaml
@@ -118,6 +118,7 @@ spec:
           args:
             - --csi-address=$(ADDRESS)
             - --default-fstype=ext4
+            - --timeout=300s
             # - --feature-gates=Topology=true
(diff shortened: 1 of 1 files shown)


assets/controller.yaml
```diff
@@ -118,6 +118,7 @@ spec:
           args:
             - --csi-address=$(ADDRESS)
             - --default-fstype=ext4
+            - --timeout=300s
```
jsafrane, line 121: I can see upstream has the same timeout in resizer and snapshotter, IMO it's worth bumping them too.

gnufied: agreed.

gnufied: fixed

PR #145, opened 2023-03-22
OCPBUGS-11146: Mark cluster un-upgradeable when CSI migration is disabled

dobsonj:
https://issues.redhat.com/browse/STOR-1269
/cc @openshift/storage

This PR depends on https://github.com/openshift/cluster-storage-operator/pull/351 for the RBAC change


dobsonj:
`W0322 02:07:04.450168       1 reflector.go:424] k8s.io/client-go@v0.26.1/tools/cache/reflector.go:169: failed to list *v1.FeatureGate: featuregates.config.openshift.io is forbidden: User "system:serviceaccount:openshift-cluster-csi-drivers:vmware-vsphere-csi-driver-operator" cannot list resource "featuregates" in API group "config.openshift.io" at the cluster scope`
I pushed an RBAC change to https://github.com/openshift/cluster-storage-operator/pull/351 to allow the operator to read featuregates.

dobsonj:
We're not adding a new FeatureSet after all, and we're not blocking upgrades.
/close

dobsonj:
Turns out we do need to block upgrades based on the API discussion. I'm going to re-use this PR, although it needs significant updates for the revised API.
/reopen

pkg/operator/vspherecontroller/checks/check_migration_disabled.go
```diff
@@ -0,0 +1,33 @@
...
+		checkResult := ClusterCheckResult{
+			CheckStatus: CheckStatusOpenshiftAPIError,
+			CheckError:  err,
+			Action:      CheckActionDegrade,
+			Reason:      fmt.Sprintf("failed to check Storage operator %s: %v", storageOperatorName, err),
+		}
+		return []ClusterCheckResult{checkResult}
+	}
+	// Migration is disabled unless this is explicitly set to CSIWithMigrationDriver
+	if storageOperator.Spec.VSphereStorageDriver != operatorv1.CSIWithMigrationDriver {
```
jsafrane, line 28: IIRC we wanted `Upgradeable=true` when `VSphereStorageDriver==""` ("I don't care").

pkg/operator/vspherecontroller/vspherecontroller.go
```diff
@@ -95,6 +96,7 @@ func NewVSphereController(
 	ocpConfigInformer := apiClients.ConfigInformers
 	configMapInformer := kubeInformers.InformersFor(cloudConfigNamespace).Core().V1().ConfigMaps()
 	infraInformer := ocpConfigInformer.Config().V1().Infrastructures()
+	storageInformer := apiClients.OCPOperatorInformers.Operator().V1().Storages()
```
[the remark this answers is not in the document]
jsafrane: And the operator does not have permissions to get/list/watch Storage CRD.

pkg/operator/vspherecontroller/checks/check_migration_disabled.go
dobsonj: Good catch, updated

pkg/operator/vspherecontroller/vspherecontroller.go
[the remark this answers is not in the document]
dobsonj: Thanks, that explains some test failures :)

pkg/operator/vspherecontroller/vsphere_environment_checker_test.go
```diff
@@ -56,10 +57,13 @@ func TestEnvironmentCheck(t *testing.T) {
...
+			storageOperator := testlib.GetStorageOperator(opv1.CSIWithMigrationDriver)
+			test.initialObjects = append(test.initialObjects, storageOperator)
+
 			stopCh := make(chan struct{})
 			defer close(stopCh)
 
-			go testlib.StartFakeInformer(commonApiClient, stopCh)
+			testlib.StartFakeInformer(commonApiClient, stopCh)
 			if err := testlib.AddInitialObjects(test.initialObjects, commonApiClient); err != nil {
 				t.Fatalf("error adding initial objects: %v", err)
```
dobsonj, line 68: There's a race condition here that didn't show up until I added a new informer. If `StartFakeInformer` is called by a go routine, the informer may not be started before the initial objects are added. Then the `Get` call fails later because it can't find the object.

jsafrane:
CI failures are actually expected until https://github.com/openshift/cluster-storage-operator/pull/351 merges.

> vmware-vsphere-csi-driver-operator" cannot list resource "storages" in API group "operator.openshift.io" at the cluster scope

dobsonj:
These _should_ pass now that https://github.com/openshift/cluster-storage-operator/pull/351 merged with the RBAC change.
/test e2e-vsphere-csi
/test e2e-vsphere
And it should merge after the e2e tests pass.
/unhold

PR #146, opened 2023-03-30
OCPBUGS-11294: Add CSI migration metric

jsafrane:
Based on https://github.com/openshift/vmware-vsphere-csi-driver-operator/pull/145

Add `vsphere_csi_migration` metric that shows value of `storage.spec.vsphereStorageDriver` as a metric.


dobsonj:
Your change looks good, but the PR shows some extra 4.14 commits compared to release-4.13.

PR #148, opened 2023-04-20
OCPBUGS-12260: Add backoff for successful storage policy creations

gnufied:
Also remove targetconfigController

Fixes https://issues.redhat.com/browse/OCPBUGS-12260


pkg/operator/storageclasscontroller/storageclasscontroller.go
```diff
@@ -86,6 +108,7 @@ func (c *StorageClassController) Sync(ctx context.Context, connection *vclib.VSp
 			utils.InstallErrorMetric.WithLabelValues(string(syncResult.CheckStatus), clusterCondition).Set(1)
 			return syncResult, checks.ClusterCheckAllGood
 		}
+		c.policyName = policyName
 
 		err := c.syncStorageClass(ctx, policyName, scState)
```
dobsonj, line 113: optional: syncStorageClass could now read from c.policyName instead of a separate argument.

pkg/operator/storageclasscontroller/storageclasscontroller.go
```diff
@@ -105,15 +128,26 @@ func (c *StorageClassController) syncStoragePolicy(ctx context.Context, connecti
 		return "", checks.MakeClusterCheckResultPass()
 	}
 
-	infra := apiDeps.GetInfrastructure()
+	// if we are running the checks after creating the policy successfully
+	// then lets run checks less frequently.
+	if !time.Now().After(c.nextCheck) && len(c.policyName) > 0 {
+		klog.Infof("Returning without running any checks")
```
dobsonj, line 134: This log statement looks like it will end up flooding the logs with this message. Could it be a debug log instead to reduce noise?

pkg/operator/storageclasscontroller/storageclasscontroller_test.go
```diff
@@ -167,6 +168,9 @@ func TestSync(t *testing.T) {
 			}
 			// err will be nil on even on failure, need to check conditions instead
```
dobsonj, line 169: Is this comment still valid?

pkg/operator/storageclasscontroller/storageclasscontroller_test.go
```diff
@@ -187,3 +191,53 @@ func TestSync(t *testing.T) {
...
+			apiDeps := getCheckAPIDependency(commonApiClient)
+			var conn *vclib.VSphereConnection
+			scController := newStorageClassController(commonApiClient, storageClass, test.storagePolicySyncFails)
+			scController.backoff = defaultBackoff
+
+			// err will be nil on even on failure, need to check conditions instead
+			err := scController.Sync(context.TODO(), conn, apiDeps)
+			if err != nil && !test.expectError {
+				t.Errorf("failed to sync controller: %+v", err)
+			}
```
dobsonj, line 230: Same question here, it's a little confusing to see a comment "err will be nil on failure" followed by checking `err != nil`.

gnufied: I have rewritten the actual test code to be better at catching error conditions.

dobsonj:
Thanks :+1: 
/lgtm

PR #152, opened 2023-04-25
OCPBUGS-12712: Add backoff for successful storage policy creations

gnufied:
Also remove targetconfigController

Fixes https://issues.redhat.com/browse/OCPBUGS-12712

### DIFF
--- a/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
+++ b/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
@@ -65,8 +65,6 @@ func NewTargetConfigController(
 		c.sync,
 	).ResyncEvery(
 		20*time.Minute, // TODO: figure out what's the idead resync time for this controller
-	).WithSyncDegradedOnError(
-		operatorClient,
 	).ToController(
 		c.name,
 		recorder.WithComponentSuffix("target-config-controller-"+strings.ToLower(name)),
--- a/pkg/operator/storageclasscontroller/storageclasscontroller.go
+++ b/pkg/operator/storageclasscontroller/storageclasscontroller.go
@@ -3,7 +3,9 @@ package storageclasscontroller
 import (
 	"context"
 	"fmt"
+	"math"
 	"strings"
+	"time"
 
 	v1 "github.com/openshift/api/config/v1"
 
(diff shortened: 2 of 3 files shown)


pkg/operator/starter.go
```diff
@@ -107,19 +106,6 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 		cloudConfigBytes,
 		controllerConfig.EventRecorder)
 
-	// targetConfigController is only used for maintaining conditions which
-	// were previously added by the operator
```
jsafrane, line 111: Why are you removing this?

gnufied: It appears to be causing unncessary churn in conditions added to the CR. I frequently see `TargetConfigControllerDegraded: False` added condition, which only gets removed later on.  So, I removed it. 

Earlier we assumed that, it is tricky to remove conditions from `Status` objects and hence we kept this controller to remove conditions from a controller, we were no longer using. But now I am thinking that, if some cluster has leftover conditions, users can remove it manually.

jsafrane: > if some cluster has leftover conditions, users can remove it manually

If that condition is Degraded=true or Available=false created by the old operator, e.g. because API server is not available during upgrade, then upgrade gets stuck.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -65,8 +65,6 @@ func NewTargetConfigController(
 		c.sync,
 	).ResyncEvery(
 		20*time.Minute, // TODO: figure out what's the idead resync time for this controller
-	).WithSyncDegradedOnError(
-		operatorClient,
```
gnufied, line 69: okay, so this is the controller code that was causing `XXXDegraded:false` condition to be added intermittently and hence I have removed it. This controller purely exists to remove older conditions anyways.

pkg/operator/starter.go
gnufied: okay, I have brought back the code and made sure it doesn't flip flop anymore.

gnufied:
Failing with "[31mERRO[0m[2023-05-04T12:32:15Z] error: Failed to acquire resource, current capacity: 0 free, 16 leased 
"

/retest

PR #155, opened 2023-05-10
STOR-1168: Bump common libraries

jsafrane:
Bump github.com/openshift/api@master github.com/openshift/build-machinery-go@master github.com/openshift/client-go@master github.com/openshift/library-go@master k8s.io/api@v0.27.1.

cc @openshift/storage


jsafrane:
/label qe-approved
discusses with QE, CI is enough.

PR #183, opened 2023-10-30
OCPBUGS-22476: Explicitly degrade the cluster when conditions are not met

gnufied:
Previously we used to rely on return values from sync function, but this creates a problem, when no checks are performed - we simply return nil in that case. This causes degraded condition to be removed from the cluster. But in-truth cluster should remain degraded.


pkg/operator/vspherecontroller/vspherecontroller.go
```diff
@@ -449,7 +450,26 @@ func (c *VSphereController) updateConditions(
 	}
 
 	updateFuncs := []v1helpers.UpdateStatusFunc{}
+
+	degradeCond := operatorapi.OperatorCondition{
+		Type:   "VMwareVSphereOperatorController" + operatorapi.OperatorStatusTypeDegraded,
```
gnufied, line 455: Note to reviewers - I am using different name than default `name + xxxxDegraded` because it appears that, that name gets clobbered by library-go `sync` function and defaults to `false`, if no error is thrown during the sync.

I made the decision to explicitly degrade the cluster (even though I had to use a different name), because otherwise we will have to store the last check result in cache somewhere (because we don't perform full checks everytime sync gets called). But in past - caching of previous check results have caused problems and I wanted code to not rely on unwanted state, if we can help it.

jsafrane: Please note about this as a huge comment into the code.

I think it's OK to add a separate Degraded condition that capture the check result, however, it's very hard for me to actually check if `updateConditions` is actually called with the right arguments in all code paths, so the condition is not accidentally cleared. IMO the operator's complexity has reached state where some refactoring / simplification is necessary (as a separate PR / techdebt epic).

jsafrane, line 455: Maybe `VMwareVSphere(Operator?)CheckDegraded` would be better, to clearly distinguish it from `VMwareVSphereControllerDegraded`

gnufied: ack

gnufied: fixed.

gnufied: Added a comment.  Usually this is the only function that updates the conditions in this controller and hence we should be okay.

PR #191, opened 2023-11-23
OCPBUGS-23342: Add dummy Progressing condition on blocked driver install

jsafrane:
When CSI driver installation is blocked, e.g. because upstream CSI driver is installed, add a dummy `Progressing` condition 
to ClusterCSIDriver. Because without any *Progressing condition, CSO will default to `Progressing=True` and the whole storage will be `Progressing=True` forever.

@openshift/storage

PR #210, opened 2024-01-15
OCPBUGS-23342: Add dummy Progressing condition on blocked driver install

jsafrane:
Manually cherry-pick https://github.com/openshift/vmware-vsphere-csi-driver-operator/pull/191 to 4.14.

When CSI driver installation is blocked, e.g. because upstream CSI driver is installed, add a dummy Progressing condition to ClusterCSIDriver. Without any `*Progressing` condition, CSO will default to `Progressing=True`.

@openshift/storage


jsafrane:
/hold
for testing

jsafrane:
/close
in favor of https://github.com/openshift/vmware-vsphere-csi-driver-operator/pull/218
That will need to be backported to 4.14 and further.

PR #218, opened 2024-01-29
OCPBUGS-27159: Mark the operator as Disabled on connection error 

jsafrane:
When the operator does not start the CSI driver, e.g. because it does not have credentials to vSphere, mark the operator as `Disabled: true`.

cluster-storage-operator will then report the whole storage as `Available: true`, `Progressing: false` with some message about the vSphere error.

Note that `Upgradeable` and `Degraded` conditions are still propagated from ClusterCSIDriver to CSO, regardless of `Disabled: true`.

Also remove `VMwareVSphereControllerAvailable` condition, it would create `Available: true` before CSI controllers had a chance to report their `Available: false`.

@openshift/storage

### DIFF
--- a/pkg/operator/vspherecontroller/vspherecontroller.go
+++ b/pkg/operator/vspherecontroller/vspherecontroller.go
@@ -7,12 +7,11 @@ import (
 	"sync"
 	"time"
 
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
-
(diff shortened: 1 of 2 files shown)


gnufied:
>Also remove VMwareVSphereControllerAvailable condition, it would create Available: true before CSI controllers had a chance to report their Available: false.

Do we have anything else that makes this `Available`? Or we don't need a `Available` condition?

PR #244, opened 2024-08-09
STOR-1819: Add manifest with OCP specific test config

jsafrane:
The LUN stress test finished under 10 minutes few times, adding 20 minutes as a buffer.

@openshift/storage


bertinatto (commented):
/lgtm

jsafrane:
/label px-approved
/label docs-approved
/label qe-approved
this is not user facing change

PR #246, opened 2024-08-13
OCPBUGS-38272: Restart CSI driver controller pod on config change

jsafrane:
Restart CSI driver controller Deployment when vsphere-csi-config-secret changes. And make a constant with the secret name.

The driver itself does not read vmware-vsphere-cloud-credentials, it's ready by the operator and translated into vsphere-csi-config-secret. So remove its secret hash from the controller.

cc @openshift/storage

### DIFF
--- a/pkg/operator/vspherecontroller/driver_starter.go
+++ b/pkg/operator/vspherecontroller/driver_starter.go
@@ -126,12 +126,12 @@ func (c *VSphereController) createCSIDriver() {
 		),
 		csidrivercontrollerservicecontroller.WithSecretHashAnnotationHook(
 			defaultNamespace,
-			cloudCredSecretName,
+			metricsCertSecretName,
 			c.apiClients.SecretInformer,
 		),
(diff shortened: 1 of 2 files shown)


jsafrane:
/cherry-pick release-4.17
before I forget about it

jsafrane:
/hold
for manual testing

pkg/operator/vspherecontroller/driver_starter.go
```diff
@@ -134,6 +134,11 @@ func (c *VSphereController) createCSIDriver() {
 			metricsCertSecretName,
 			c.apiClients.SecretInformer,
 		),
+		csidrivercontrollerservicecontroller.WithSecretHashAnnotationHook(
+			defaultNamespace,
+			driverConfigSecretName,
```
jsafrane, line 139: @bertinatto I had to fix the order of namespace / secret name, it's different than in `WithSecretDaemonSetAnnotationHook`
PTAL again

/hold cancel

bertinatto (commented):
/lgtm

PR #248, opened 2024-08-30
OCPBUGS-38274: Remove conditions and controllers

gnufied:
Remove stuff statically


pkg/operator/vspherecontroller/operator_removal.go
```diff
@@ -0,0 +1,135 @@
...
+func (c *VSphereController) addDisabledCondition() v1helpers.UpdateStatusFunc {
+	return func(status *operatorapi.OperatorStatus) error {
+		disabledConditionName := c.getDisabledConditionName()
+		disabledCond := operatorapi.OperatorCondition{
+			Type:    disabledConditionName,
+			Status:  operatorapi.ConditionTrue,
+			Message: "Operator is disabled",
+		}
+		v1helpers.SetOperatorCondition(&status.Conditions, disabledCond)
+		return nil
```
gnufied, line 133: Usually this gets removed once - https://github.com/openshift/vmware-vsphere-csi-driver-operator/blob/master/pkg/operator/vspherecontroller/vspherecontroller.go#L579 code runs, but it appears to be buggy and sometimes this condition is not removed. 

I can reproduce this via e2e. :(

Makefile
```diff
@@ -40,3 +40,11 @@ test-e2e:
...
 
 .PHONY: test-e2e
+
+operator-e2e-test: GO_TEST_PACKAGES :=./test/e2e/...
+operator-e2e-test: GO_TEST_FLAGS += -v
+operator-e2e-test: GO_TEST_FLAGS += -count=1
+operator-e2e-test: GO_TEST_FLAGS += -timeout 3h
+operator-e2e-test: GO_TEST_FLAGS += -p 1
+operator-e2e-test: test-unit
+.PHONY: operator-e2e-test
```
mpatlasov, line 50: Do we need some automation to make this target regularly? (If some innocent code change breaks the test in future, how we'll discover that the test got broken?)

test/e2e/operator_removal_test.go
```diff
@@ -0,0 +1,238 @@
...
+	err := makeClusterCSIDriverManaged(context.TODO(), client)
+	if err != nil {
+		t.Fatalf("Failed to update ClusterCSIDriver: %v", err)
+	}
+	waitForStorageAvailable(t, configClient, client, kubeClient)
+}
+
+func waitForStorageResourceRemoval(t *testing.T, client *operatorclient.Clientset, kubeClient *kubernetes.Clientset) error {
+	// make sure storage to be removed
+	klog.Infof("Waiting for storage resource to be removed")
```
mpatlasov, line 87: Should we use `t.Log` here to be consistent with all other t.Log-s in the test?

test/library/client.go
```diff
@@ -0,0 +1,20 @@
...
+	"k8s.io/client-go/tools/clientcmd/api"
+)
+
+// NewClientConfigForTest returns a config configured to connect to the api server
+func NewClientConfigForTest() (*rest.Config, error) {
+	loader := clientcmd.NewDefaultClientConfigLoadingRules()
+	clientConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loader, &clientcmd.ConfigOverrides{ClusterInfo: api.Cluster{InsecureSkipTLSVerify: true}})
+	config, err := clientConfig.ClientConfig()
+	if err == nil {
+		fmt.Printf("Found configuration for host %v.\n", config.Host)
```
mpatlasov, line 17: Should we use `t.Log` here to be consistent with all other t.Log-s in the test?

pkg/operator/storageclasscontroller/storageclasscontroller.go
```diff
@@ -145,6 +147,20 @@ func (c *StorageClassController) Sync(ctx context.Context, connections []*vclib.
...
 }
 
+func (c *StorageClassController) SyncRemove(ctx context.Context) error {
+	scString := string(c.manifest)
+	pairs := []string{
+		"${STORAGE_POLICY_NAME}", "sample",
+	}
+
+	policyReplacer := strings.NewReplacer(pairs...)
+	scString = policyReplacer.Replace(scString)
```
mpatlasov, line 157: In my experiments `resourceapply.DeleteStorageClass()` below works equally fine if we replace `${STORAGE_POLICY_NAME}` or don't. Can we remove L152-157 altogether?

pkg/operator/vspherecontroller/vspherecontroller.go
```diff
@@ -160,11 +164,6 @@ func (c *VSphereController) sync(ctx context.Context, syncContext factory.SyncCo
 		return err
 	}
 
-	if opSpec.ManagementState != operatorapi.Managed {
-		klog.Warningf("%s: ManagementState is %s, skipping", c.name, opSpec.ManagementState)
-		return nil
-	}
-
```
mpatlasov, line 167: This essentially moves the check from the beginning of `sync()` to the place after infra checks. Is this change relevant to the bug (OCPBUGS-38274) or it's piggybacked? As for me, I'd vote for the previous position of the check: if state!=Managed, we shouldn't care about infra from inside vsphere storage operator.

Makefile
gnufied: I added - https://github.com/openshift/release/pull/56369

pkg/operator/vspherecontroller/operator_removal.go
```diff
@@ -0,0 +1,135 @@
...
+		"webhook/pdb.yaml",
+		// dynamic assets used by controller
+		"volumesnapshotclass.yaml",
+		"controller.yaml",
+		"node.yaml",
+		"servicemonitor.yaml",
+		"webhook/deployment.yaml",
+		"vsphere_cloud_config_secret.yaml",
+		"vsphere_features_config.yaml",
+	}
```
mpatlasov, line 73: This code-dup is error-prone. Any time we'll add new resource in driver_starter.go or starter.go, it is easy to overlook that this list must be updated. It would be cleaner to define proper lists of yaml-s in a single yaml_resources.go, and then re-use them here and where we install resources. (we can do it later, as a cleanup patch, of course).

pkg/operator/vspherecontroller/operator_removal.go
```diff
@@ -0,0 +1,135 @@
...
+		// Static assets used by the webhook
+		"webhook/config.yaml",
+		"webhook/sa.yaml",
+		"webhook/service.yaml",
+		"webhook/configuration.yaml",
+		"webhook/rbac/role.yaml",
+		"webhook/rbac/rolebinding.yaml",
+		"webhook/pdb.yaml",
+		// dynamic assets used by controller
+		"volumesnapshotclass.yaml",
```
mpatlasov, line 66: volumesnapshotclass.yaml is static asset, isn't it?

test/e2e/operator_removal_test.go
```diff
@@ -0,0 +1,238 @@
...
+	operatorEnabled
+	clusterCSIDriverNotFound
+)
+
+var (
+	WaitPollInterval = time.Second
+	WaitPollTimeout  = 10 * time.Minute
+)
+
+func TestStorageRemoval(t *testing.T) {
```
mpatlasov, line 47: Can we add explicit comment before `func TestStorageRemoval` that this test does not check if any of static resources or conditions removed? I understand it would be tedious to check all resources and conditions, but may be someone would desire to increase coverage adding them in future.

pkg/operator/vspherecontroller/vspherecontroller.go
gnufied: Yeah - this change is intentional. I don't want to be removing stuff if platform is not of `vSphere` type. Though I agree - I do not see how this operator will get "activated" if platform is not vsphere. But - this is more like another failsafe.

pkg/operator/vspherecontroller/operator_removal.go
gnufied: Yes - over time I expect that we will update the e2e to ensure that everything that this operator installs will be removed.  Also - it is a good idea to define single source of truth for YAML files, which can be reused both for creation and deletion.  Lets do that as a follow up.

pkg/operator/vspherecontroller/operator_removal.go
gnufied: So, although it is a static asset, it only gets installed if certain CRD is present.  So I have kept it under dynamic assets.

test/e2e/operator_removal_test.go
gnufied: added

pkg/operator/storageclasscontroller/storageclasscontroller.go
gnufied: fixed.

PR #251, opened 2024-09-10
OCPBUGS-41532: Add check for vCenter version

gnufied:
Fixes https://issues.redhat.com/browse/OCPBUGS-41532?filter=12399251

### DIFF
--- a/pkg/cnsmigration/migrator.go
+++ b/pkg/cnsmigration/migrator.go
@@ -352,7 +352,7 @@ func (c *CNSVolumeMigrator) checkRequiredVersion() error {
 	}
 	match, message, err := utils.CheckForMinimumPatchedVersion(minRequirement, apiVersion, build)
 	if err != nil {
-		return fmt.Errorf("error checking for minimum version: %v", err)
+		return fmt.Errorf("cns-migration requires %s, error checking for minimum version: %v", message, err)
 	}
 	if !match {
 		return fmt.Errorf("vCenter version %s:%s is not supported: minimum supported version is - %s", apiVersion, build, message)
--- a/pkg/cnsmigration/migrator.go
+++ b/pkg/cnsmigration/migrator.go
@@ -88,17 +88,13 @@ func (c *CNSVolumeMigrator) StartMigration(ctx context.Context, volumeFile strin
 	}
 
 	// check if the vCenter version is supported
-	supported, err := c.checkRequiredVersion()
+	err = c.checkRequiredVersion()
 	if err != nil {
-		printError("error checking for required version: %v", err)
+		printError("error checking minimum version of vCenter that supports volume migration: %v", err)
 		return err
 	}
-	if !supported {
-		printError("vCenter version is not supported")
-		return fmt.Errorf("vCenter version is not supported")
-	} else {
-		printGreenInfo("vCenter version supports CNS volume Migration")
-	}
+
+	printGreenInfo("vCenter version supports CNS volume Migration")
 
 	printGreenInfo("logging successfully to vcenter")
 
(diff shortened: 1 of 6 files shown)


pkg/cnsmigration/migrator.go
```diff
@@ -323,6 +342,20 @@ func (c *CNSVolumeMigrator) loginToVCenter(ctx context.Context) error {
...
+	if err != nil {
+		return fmt.Errorf("error getting vCenter version: %v", err)
+	}
+	match, message, err := utils.CheckForMinimumPatchedVersion(minRequirement, apiVersion, build)
+	if err != nil {
+		return fmt.Errorf("error checking for minimum version: %v", err)
+	}
+	if !match {
+		return fmt.Errorf("vCenter version %s:%s is not supported: minimum supported version is - %s", apiVersion, build, message)
+	}
```
jsafrane, line 359: nit: can `CheckForMinimumPatchedVersion` return error when the version don't match? I don't think having a separate `match bool` is very useful. `match=false` has the same consequence as `err = nil` - both return error to the caller.

pkg/operator/utils/version_check_test.go
```diff
@@ -0,0 +1,71 @@
...
+			vSphereVersion:   "7.0.3",
+			buildNumber:      "23788037",
+			meetsRequirement: true,
+		},
+		{
+			name:             "when vSphere version meets minimum 8 series requirement",
+			vSphereVersion:   "8.0.2",
+			buildNumber:      "23504391",
+			meetsRequirement: true,
+		},
```
jsafrane, line 25: just for the fun, please add a test with v9. I think it will work just fine.

gnufied: it depends on build number we specify for 9 series in current code. It will not pass for any random build number fwiw. :/

pkg/cnsmigration/migrator.go
gnufied: So I have enhanced some of the code to return and print actual required version if parsing etc also fails. IMO - this will lead to better error reporting than the alternative.

pkg/operator/utils/version_check.go
```diff
@@ -0,0 +1,87 @@
...
+	MinimumVersion7Series string
+	MinimumBuild7Series   int
+
+	MinimumVersion8Series string
+	MinimumBuild8Series   int
+}
+
+func CheckForMinimumPatchedVersion(minRequirement PatchVersionRequirements, vSphereVersion string, build string) (bool, string, error) {
+	hasMinimumApiVersion, err := IsMinimumVersion(minRequirement.MinimumVersion7Series, vSphereVersion)
+	min7SeriesString := fmt.Sprintf(">= %s-%d", minRequirement.MinimumVersion7Series, minRequirement.MinimumBuild7Series)
```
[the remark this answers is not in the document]
gnufied: fixed.

pkg/operator/utils/version_check_test.go
jsafrane: So, when v9 is introduced, we need to update the code to add support for it? That's unfortunate.

pkg/operator/utils/version_check.go
```diff
@@ -0,0 +1,87 @@
...
+
+type PatchVersionRequirements struct {
+	MinimumVersion7Series string
+	MinimumBuild7Series   int
+
+	MinimumVersion8Series string
+	MinimumBuild8Series   int
+}
+
+func CheckForMinimumPatchedVersion(minRequirement PatchVersionRequirements, vSphereVersion string, build string) (bool, string, error) {
```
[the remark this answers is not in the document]
jsafrane: the if/else branches share a non-trivial amount of code, can you please refactor it a bit?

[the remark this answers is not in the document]
gnufied: fixed.

pkg/operator/utils/version_check_test.go
gnufied: I have added code with the assumption that anything higher than 8 will automatically be supported.

PR #252, opened 2024-09-13
NO-JIRA: Add docs for cns volume migration

gnufied:


docs/cns-migration.md
```diff
@@ -0,0 +1,49 @@
...
+
+```oc get pv -o json|jq -r '.items[] | select(.spec.csi.driver=="csi.vsphere.vmware.com")| .metadata.name'```
+
+
+### Now all PVs in the aforementioned file can be migrated via:
+
+```
+export KUBECONFIG=<path-to-kube-config>
+
+./cns-migration --source <source_datastore> --destination <destination_datastore> -volume-file <path_to_file_with_pv_names>```
```
dobsonj, line 45: The ending ``` needs to be on a new line, otherwise the preformatted block lasts until EOF.

gnufied: fixed!

PR #258, opened 2024-09-19
OCPBUGS-38368: Fix the config loading warning

rhrmo:
Fixes https://issues.redhat.com/browse/OCPBUGS-38368

### DIFF
--- a/third_party/cloud-provider-vsphere/pkg/common/config/config_yaml.go
+++ b/third_party/cloud-provider-vsphere/pkg/common/config/config_yaml.go
@@ -205,7 +205,6 @@ func ReadRawConfigYAML(byConfig []byte) (*CommonConfigYAML, error) {
 	}
 
 	if err := yaml.Unmarshal(byConfig, &cfg); err != nil {
-		klog.Errorf("Unmarshal failed: %s", err)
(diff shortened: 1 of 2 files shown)


third_party/cloud-provider-vsphere/pkg/common/config/config_yaml.go
```diff
@@ -205,7 +205,6 @@ func ReadRawConfigYAML(byConfig []byte) (*CommonConfigYAML, error) {
 	}
 
 	if err := yaml.Unmarshal(byConfig, &cfg); err != nil {
-		klog.Errorf("Unmarshal failed: %s", err)
```
rhrmo, line 208: I do not think we need this logged, if INI passes we don't need to know yaml unmarshal failed and if it doesn't then we would have the same error message twice in the logs.

vr4manta:
The intention of this third party code was to try to work around output from upstream.  The changes in this PR fix looks fine.  The third party override code will get removed when the following PR gets fixed:
https://github.com/kubernetes/cloud-provider-vsphere/pull/1225

gnufied:
Can you backport this to 4.17?

PR #272, opened 2024-11-15
OCPBUGS-44431: Fix link to migration documentation

tsmetana:
The old KCS article has disappeared: replace it with the link to the OpenShift documentation in the log message.


gnufied:
So, this just links to a document which links to a KCS that is missing. :-) 
/hold

tsmetana:
This is now fixed on the server side -- the old article is accessible again. This PR is not needed.

PR #287, opened 2025-02-04
STOR-2141: add support for maxAllowedBlockVolumesPerNode

RomanBednar:
## Depends on

- openshift/api: https://github.com/openshift/api/pull/2190
- openshift/client-go: https://github.com/openshift/client-go/pull/313

## Manual verification

Test value limits for `maxAllowedBlockVolumesPerNode` field:
```
oc patch clustercsidriver/csi.vsphere.vmware.com --type=merge -p '{"spec":{"driverConfig":{"vSphere":{"maxAllowedBlockVolumesPerNode":-1}}}}'
The ClusterCSIDriver "csi.vsphere.vmware.com" is invalid: spec.driverConfig.vSphere.maxAllowedBlockVolumesPerNode: Invalid value: -1: spec.driverConfig.vSphere.maxAllowedBlockVolumesPerNode in body should be greater than or equal to 1

$ oc patch clustercsidriver/csi.vsphere.vmware.com --type=merge -p '{"spec":{"driverConfig":{"vSphere":{"maxAllowedBlockVolumesPerNode":0}}}}'
The ClusterCSIDriver "csi.vsphere.vmware.com" is invalid: spec.driverConfig.vSphere.maxAllowedBlockVolumesPerNode: Invalid value: 0: spec.driverConfig.vSphere.maxAllowedBlockVolumesPerNode in body should be greater than or equal to 1

$ oc patch clustercsidriver/csi.vsphere.vmware.com --type=merge -p '{"spec":{"driverConfig":{"vSphere":{"maxAllowedBlockVolumesPerNode":256}}}}'
The ClusterCSIDriver "csi.vsphere.vmware.com" is invalid: spec.driverConfig.vSphere.maxAllowedBlockVolumesPerNode: Invalid value: 256: spec.driverConfig.vSphere.maxAllowedBlockVolumesPerNode in body should be less than or equal to 255
```

Validate `maxAllowedBlockVolumesPerNode` value propagation to driver deployment as `MAX_VOLUMES_PER_NODE`:
```
$ oc patch clustercsidriver/csi.vsphere.vmware.com --type=merge -p '{"spec":{"driverConfig":{"vSphere":{"maxAllowedBlockVolumesPerNode":60}}}}'
clustercsidriver.operator.openshift.io/csi.vsphere.vmware.com patched

$ oc -n openshift-cluster-csi-drivers get deployment.apps/vmware-vsphere-csi-driver-controller -o jsonpath='{.spec.template.spec.containers[0].env}'
[{"name":"CSI_ENDPOINT","value":"unix:///var/lib/csi/sockets/pluginproxy/csi.sock"},{"name":"X_CSI_MODE","value":"controller"},{"name":"VSPHERE_CSI_CONFIG","value":"/etc/kubernetes/vsphere-csi-config/cloud.conf"},{"name":"INCLUSTER_CLIENT_QPS","value":"100"},{"name":"INCLUSTER_CLIENT_BURST","value":"100"},{"name":"CSI_NAMESPACE","valueFrom":{"fieldRef":{"apiVersion":"v1","fieldPath":"metadata.namespace"}}},{"name":"X_CSI_SERIAL_VOL_ACCESS_TIMEOUT","value":"3m"},{"name":"X_CSI_SPEC_DISABLE_LEN_CHECK","value":"true"},{"name":"MAX_VOLUMES_PER_NODE","value":"60"}]

$ oc -n openshift-cluster-csi-drivers get daemonset.apps/vmware-vsphere-csi-driver-node -o jsonpath='{.spec.template.spec.containers[0].env}'
[{"name":"NODE_NAME","valueFrom":{"fieldRef":{"apiVersion":"v1","fieldPath":"spec.nodeName"}}},{"name":"CSI_ENDPOINT","value":"unix:///csi/csi.sock"},{"name":"X_CSI_MODE","value":"node"},{"name":"X_CSI_SPEC_DISABLE_LEN_CHECK","value":"true"},{"name":"CSI_NAMESPACE","valueFrom":{"fieldRef":{"apiVersion":"v1","fieldPath":"metadata.namespace"}}},{"name":"MAX_VOLUMES_PER_NODE","value":"60"}]
```

Validate propagation to CSINode as allocatable count:
```
$ oc get csinode/ci-ln-k30mn5t-c1627-2tk2k-worker-0-72mfn -o jsonpath='{.spec.drivers[0].allocatable.count}'
60
```

If `maxAllowedBlockVolumesPerNode` is unset (for example after cluster upgrade) we must use default value (never zero):

```
$ oc get clustercsidriver/csi.vsphere.vmware.com -o jsonpath='{.spec.driverConfig}'
{"driverType":""}
$ oc -n openshift-cluster-csi-drivers get deployment.apps/vmware-vsphere-csi-driver-controller -o jsonpath='{.spec.template.spec.containers[0].env[?(@.name=="MAX_VOLUMES_PER_NODE")]}'
{"name":"MAX_VOLUMES_PER_NODE","value":"59"}
```


assets/vsphere_features_config.yaml
```diff
@@ -17,6 +17,7 @@ data:
   "improved-volume-topology": "false" # operator will set this value to true if cluster supports CSI topology
   "improved-csi-idempotency": "true" # Causes operator to create CRs for volume operations for sync them with k8s server
   "multi-vcenter-csi-topology": "true" # add support multiple vcenters. Although we don't support this feature, having it disabled causes driver to crash loop.
+  "max-pvscsi-targets-per-vm": "true" # enables higher limit (255) in the driver for volumes per node in vSphere 8, the limit value is set by `MAX_VOLUMES_PER_NODE`
```
gnufied, line 20: This is already enabled, so setting this doesn't do anything fwiw.

RomanBednar: I see now, thank you! Explained here: https://github.com/openshift/enhancements/pull/1748#discussion_r1943511577

RomanBednar:
/assign @gnufied 

For review - feel free to reassign to other candidate.

pkg/operator/vspherecontroller/checks/check_nodes.go
```diff
@@ -242,6 +261,23 @@ func (n *NodeChecker) getHost(ctx context.Context, checkOpts CheckArgs, hostRef
...
+
+	// Get ClusterCSIDriver object
+	clusterCSIDriver, err := n.ClusterCSIDriverLister.Get(utils.VSphereDriverName)
+	if err != nil {
+		return 0, fmt.Errorf("failed to get ClusterCSIDriver: %v", err)
+	}
+
+	// Check if a custom value is set in the ClusterCSIDriver
+	if clusterCSIDriver != nil && clusterCSIDriver.Spec.DriverConfig.VSphere != nil {
+		maxVolumesPerNode = int(clusterCSIDriver.Spec.DriverConfig.VSphere.MaxAllowedBlockVolumesPerNode)
```
gnufied, line 275: What does this return when no limit is set? 0 ??

gnufied:
Shouldn't all this code be behind a feature gate?

pkg/operator/vspherecontroller/checks/check_nodes.go
RomanBednar: Yeah, this needs one more check - added.

RomanBednar:
> Shouldn't all this code be behind a feature gate?

It could, but do we want it? I thought we don't because:
- we did not featuregate the code for snapshot options a while ago, and this is very similar
- we have the API behind featuregate and we decided to let operator set default value, that means if the new field is unset we apply the default (59)

So I'm not sure what exactly we would gain from featuregating this code, but might have missed something so please share your thoughts.

gnufied:
>So I'm not sure what exactly we would gain from featuregating this code, but might have missed something so please share your thoughts.

But then, why did we introduce https://github.com/openshift/api/blob/master/features/features.go#L821 featuregate in first place? Isn't the idea that, user must first enable that featuregate before this can be used?

If snapshot feature went TP without a featuregate check in the code, that was a mistake. cc @jsafrane 

There are other reasons why this should be featuregated, say I am a user who discovers this feature, which is available by default (and I didn't read docs), then if this feature is removed in next release, my cluster will be broken. There is also risk of code change breaking stable features (no matter how small that chance is).

RomanBednar:
>But then, why did we introduce https://github.com/openshift/api/blob/master/features/features.go#L821 featuregate in first place?

That feature gate now affects which CRDs are applied, so unless users enable that feature gate they will not be able to set the field (it will appear as invalid for clustercsidriver). 

>There is also risk of code change breaking stable features (no matter how small that chance is).

This is a valid point, but if there's a GA'd feature, it must have e2e tests, so in theory we should see any breakage well. But this was historically not always the case I think. 

> if this feature is removed in next release, my cluster will be broken

What scenario are we talking about exactly? I can see removal as problematic, either because of the field presence (we can remove it by operator) or having more volumes than 59 attached already (we won't be able to deal with this anyway). But feature removal should be only for cases like vmware removing it first right? And we don't know when/if that happens and we can't keep featuregates forever I believe.

So what's the suggestion here exactly?

pkg/operator/vspherecontroller/driver_starter.go
```diff
@@ -324,3 +330,65 @@ func getOperatorSyncState(operatorClient v1helpers.OperatorClientWithFinalizers)
...
+		if err != nil {
+			return err
+		}
+		ds.Spec.Template.Spec.Containers = containers
+		return nil
+	}
+}
+
+// WithMaxVolumesPerNodeDeploymentHook sets the MAX_VOLUMES_PER_NODE environment variable in the Deployment container specifications.
+func WithMaxVolumesPerNodeDeploymentHook(clusterCSIDriverLister operatorlistersv1.ClusterCSIDriverLister, featureGate featuregates.FeatureGate) deploymentcontroller.DeploymentHookFunc {
```
gnufied, line 385: I didn't think this was required in control-plane deployment.

pkg/operator/vspherecontroller/vsphere_environment_checker.go
```diff
@@ -37,15 +38,17 @@ type vSphereEnvironmentCheckerComposite struct {
...
 		backoff:   defaultBackoff,
 		nextCheck: time.Now(),
 	}
 	checker.checkers = []checks.CheckInterface{
 		&checks.CheckExistingDriver{},
 		&checks.VCenterChecker{},
-		&checks.NodeChecker{},
+		&checks.NodeChecker{
+			ClusterCSIDriverLister: clusterCSIDriverLister,
+		},
```
gnufied, line 51: So, rather than passing this as one-off, shouldn't this be part of - https://github.com/openshift/vmware-vsphere-csi-driver-operator/blob/main/pkg/operator/vspherecontroller/checks/api_interface.go#L27 ? That will keep everything uniform between checkers.

pkg/operator/vspherecontroller/checks/check_nodes.go
```diff
@@ -169,6 +172,25 @@ func (n *NodeChecker) checkOnNode(workInfo nodeChannelWorkData) ClusterCheckResu
 		return MakeClusterUnupgradeableError(CheckStatusDeprecatedESXIVersion, reason)
 	}
 
+	// Check for maxAllowedBlockVolumesPerNode and degrade if host is not on minimal required ESXI version
+	if workInfo.checkOpts.featureGate.Enabled(features.FeatureGateVSphereConfigurableMaxAllowedBlockVolumesPerNode) {
```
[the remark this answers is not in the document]
RomanBednar: We're talking about two different scenarios here, what I meant was making sure we don't break clusters if we ever need to disable the feature. Something like dropping disabled fields in k/k upstream API strategy, example: https://github.com/kubernetes/kubernetes/blob/195803cde570ad1025a78e36cdbef76bddbc4c33/pkg/api/persistentvolumeclaim/util.go#L101-L109

Now we have the enablement covered by feature gate and CRDs applied which is sufficient for now, so keeping it as it is.

pkg/operator/vspherecontroller/driver_starter.go
RomanBednar: That's what I thought as well but it was added to assets here: https://github.com/openshift/vmware-vsphere-csi-driver-operator/commit/7cf46b969cc1329e9d09b855747ee0d1f9bc6a76

And `CSINode` is created for master nodes too which is where the volume limit is reflected as:
```
spec:
  drivers:
  - allocatable:
      count: 59
```

But the reasons are not entirely clear to me, maybe consistency? Or because we can't prevent workloads being scheduled on masters?

Anyway, if you think this should be removed we shouldn't do it in scope of this PR as it would be unrelated change.

pkg/operator/vspherecontroller/vsphere_environment_checker.go
RomanBednar: Right, it should - changed.

pkg/operator/vspherecontroller/vsphere_environment_checker_test.go
```diff
@@ -94,7 +94,7 @@ func TestEnvironmentCheck(t *testing.T) {
 			workingDir, _ := os.Getwd()
 			t.Logf("working directory is: %s", workingDir)
 
-			checker := newVSphereEnvironmentChecker()
+			checker := newVSphereEnvironmentChecker(commonApiClient.ClusterCSIDriverInformer.Lister())
```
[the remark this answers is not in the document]
RomanBednar: Changed.

pkg/operator/utils/config.go
```diff
@@ -12,6 +15,18 @@ func addOption(options map[string]string, name string, intValue *uint32) {
...
 
+func GetMaxVolumesPerNode(clusterCSIDriver *opv1.ClusterCSIDriver, featureGate featuregates.FeatureGate) int {
+	maxVolumesPerNode := MaxVolumesPerNodeVSphere7
+
+	if featureGate != nil && featureGate.Enabled(features.FeatureGateVSphereConfigurableMaxAllowedBlockVolumesPerNode) &&
+		clusterCSIDriver != nil && clusterCSIDriver.Spec.DriverConfig.VSphere != nil &&
+		clusterCSIDriver.Spec.DriverConfig.VSphere.MaxAllowedBlockVolumesPerNode > 0 {
+		maxVolumesPerNode = int(clusterCSIDriver.Spec.DriverConfig.VSphere.MaxAllowedBlockVolumesPerNode)
+	}
+
```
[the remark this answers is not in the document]
RomanBednar: That's an option too, but caller in this context you mean `setMaxVolumesPerNodeEnvVar`? I think it would make more sense to check featuregate in the hook func.

pkg/operator/vspherecontroller/driver_starter.go
gnufied: The allocatable limit from control-plane drivers don't make any sense. The master nodes actually should have 2 drivers running. One in worker mode and another one in control-plane mode. The driver that runs with controller doesn't have any impact on reported limit.

To be clear - I am *not* asking you to remove the limit from asset file. I am just asking you to not add this hook at all here. Just leave it as it is. Less code the better.

RomanBednar: Alright, removed.

gnufied:
hmm, looks like vsphere operator tests that check storage removal are failing. Is that a real failure? @RomanBednar can you check?

RomanBednar:
@gnufied Looks like a flake, green on next try.

PR #302, opened 2025-03-27
STOR-2245: add manifest-topology.yaml for topology related feature and test

duanwei33:
When adding the multi-zone/multi-vcenter CI, some cases will check if there are at least 2 nodes in the same zone and need the topology Capability. 
The other profile that doesn't have the zonal configuration will still need the previous manifest. (The topology key is null)


jsafrane:
/label acknowledge-critical-fixes-only
this only adds a manifest, it's not used yet.

PR #312, opened 2025-10-21
OCPBUGS-62344: Fix escaping of password

gnufied:
This improves escaping of golang escape characters using json library.

Also serializing the ini file using `iniv1` library appears to be incompatible with gcfg for various formats.  We change the operator to use a handrolled version.  We test if result is parseable by gcfg.

All of existing tests are still passing and this appears to fix whole host of character escaping issues, so I think we should be good.

### DIFF
--- a/pkg/operator/vspherecontroller/vspherecontroller.go
+++ b/pkg/operator/vspherecontroller/vspherecontroller.go
@@ -1,9 +1,7 @@
 package vspherecontroller
 
 import (
-	"bytes"
 	"context"
-	"encoding/json"
 	"fmt"
 	"os"
 	"regexp"
@@ -88,6 +86,8 @@ const (
 	storageClassControllerName        = "VMwareVSphereDriverStorageClassController"
 	storageClassName                  = "thin-csi"
 )
+
+var reEscape = regexp.MustCompile(`["\\]`)
 
 type conditionalControllerInterface interface {
 	Run(ctx context.Context, workers int)
@@ -730,10 +730,7 @@ func (c *VSphereController) applyClusterCSIDriverChange(
 			return nil, fmt.Errorf("error getting user and password: %v", err)
 		}
 
-		escapedUser, escapedPassword, err := escapeUserAndPassword(user, password)
-		if err != nil {
-			return nil, err
-		}
+		escapedUser, escapedPassword := escapeUserAndPassword(user, password)
 
 		for pattern, value := range map[string]string{
 			"${VCENTER}":     vcenter.VCenterIP,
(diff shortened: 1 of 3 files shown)


pkg/operator/vspherecontroller/vspherecontroller_test.go
```diff
@@ -1188,7 +1281,7 @@ func TestEscapeQuotesAndBackslashes(t *testing.T) {
 		{
 			name:     "Already escaped",
 			input:    `Pass\"word\\1234\"`,
-			expected: `Pass\\\"word\\\\1234\\\"`,
+			expected: `Pass\"word\\1234\"`,
```
jsafrane: This is a change of behavior. User provided password with `\"` and `\\`, yet the operator creates an ini file whose password will have `"` and `\`.

pkg/operator/vspherecontroller/vspherecontroller.go
```diff
@@ -818,26 +819,64 @@ func getUserAndPassword(namespace string, secretName string, vcenter string, inf
...
 // escapeQuotesAndBackslashes escapes double quotes and backslashes in the input string.
-func escapeQuotesAndBackslashes(input string) string {
-	return reEscape.ReplaceAllString(input, `\$0`)
+func escapeQuotesAndBackslashes(data string) (string, error) {
+	var js json.RawMessage
+	data = strings.ReplaceAll(data, "\n", " ")
+	jsonString := createJsonString(data)
+	byteValue := []byte(jsonString)
+	//escape special characters only if json unmarshall results in an error
+	if err := json.Unmarshal(byteValue, &js); err != nil {
```
jsafrane: What's the purpose of this?  Consider a password that resembles chunk of json, like `", "key2": "value2`. This check will consider it as a password that does not need escaping. But it does!

pkg/operator/vspherecontroller/vspherecontroller.go
```diff
@@ -818,26 +819,64 @@ func getUserAndPassword(namespace string, secretName string, vcenter string, inf
...
+		var buf bytes.Buffer
+		encoder := json.NewEncoder(&buf)
+		// Disable HTML escaping to avoid conversion of &, >, < etc to unicode char sequences
+		encoder.SetEscapeHTML(false)
+		// encoder.Encode will escape special characters
+		err = encoder.Encode(data)
```
jsafrane: I checked [JSON escaping](https://cs.opensource.google/go/go/+/master:src/encoding/json/encode.go;l=1020) and [gcfg escaping](https://github.com/openshift/vmware-vsphere-csi-driver-operator/blob/0d9a69f3e60f827e083934cc85141b7b1b4d1aa7/vendor/gopkg.in/gcfg.v1/read.go#L16), they're different when handling ASCII characters <32. I _think_ we can live with that - `\r`, `\b` or `\f` in a password?

Still, there should be at least a comment about that.

